PR #65, opened 2020-06-18
Wednesday's Work

marcfishelquantil:


docs/portal/edge-configurations/testing-property.md
```diff
@@ -26,7 +26,7 @@ faster.mydomain.com CNAME staging.qtlgslb.com
 ```
 
 If you use a smart DNS service, you can also direct a small portion of the production traffic to the staging hostname to have a more thorough test.
-- Modify the ```/etc/hosts``` file to hard code the hostname to be accelerated to one of the staging servers IP addresses. For example:
+- Modify the ```/etc/hosts``` file to hard code the hostname to be accelerated to one of the staging server's IP addresses. For example:
```
burtonquantil, line 29: Are you sure this shouldn't be "servers'" ?

PR #85, opened 2020-08-07
NCAS-2034 #comment updated to the latest which describes additional s…

burtonquantil:
…ervice quota APIs. It is  needs to be reviewed by our QA team before merging to production though. It should also fix NGDOC-52.


nibinqtl:
Be careful with the conflicts with other PRs.

PR #89, opened 2020-08-18
Merge latest updates to master

nibinqtl:


docs/edge-logic/supported-directives.md
```diff
@@ -596,11 +596,22 @@ Assigns a value to the specified variable. No change to the public version. In p
...
 
-**Contexts:** http, server
+**Contexts:** http, server, ~~location~~
 
 
-Sets the size of the slice when fetching large files from the origin. <span style='background:#ffff06'>We made a change that disallows this directive in any "location" block.</span> If slice is enabled, the entire website must use the same slice size. This behavior avoids potential problems when trying to follow origin’s redirection. The valid values are 0, which disables slicing, OR an [nginx size](http://nginx.org/en/docs/syntax.html) that is no less than `512k` and up to `512m`.
-For this directive to work correctly, the origin should support HTTP status 206 for ranged requests.
+Sets the size of the slices when fetching large files from the origin. The valid values are 0, which disables slicing, OR an [nginx size](http://nginx.org/en/docs/syntax.html) that is no less than `512k` and up to `512m`. The origin has to support ranged request and respond with status code 206. When caching is desired, the statement `proxy_cache_valid 206 ...` should be used to enable caching of the ranged responses. We made the following changes to this directive on top of the open-source version:
```
burtonquantil, line 602: 1) Is "ranged request" a standard term?  I have only seen the term "range request". 
2)  You suggest using "proxy_cache_valid".  However, our documentation for that mentions ETA of  Oct 2020.
How is a customer supposed to use something that is not ready?

docs/edge-logic/supported-directives.md
```diff
@@ -596,11 +596,22 @@ Assigns a value to the specified variable. No change to the public version. In p
...
+* We disallowed this directive in any "location" block. As a result, the entire website must use the same slice size. This is to avoid potential problems when a request needs to be processed in multiple locations.
+* CDN360 requires all cached slices carry the same ETag value to ensure the content is consistent. When a slice fetched from the origin has a value that is different from the cached ones, any in-progress transfers to clients are terminated and all the cached slices are purged immediately. So please make sure the ETag of a file on origin does not change unless the file content really changes. This behavior can be disabled by `slice_ignore_etag on;`.
```
burtonquantil, line 604: 1) "So please make sure...." is not a proper English sentence. 
Suggestion:  Please make sure the ETag value of each file on the origin does not change unless the file's content has changed.
2) I prefer "disabled using"  to "disabled by"

docs/edge-logic/supported-directives.md
```diff
@@ -596,11 +596,22 @@ Assigns a value to the specified variable. No change to the public version. In p
...
+
+### `slice_ignore_etag`
+
+<span class="badge">standard</span> <span class="badge primary">CDN360 Proprietary</span>
+
+**Syntax:** `slice_ignore_etag on/off;` <br/>
+**Default:** `slice_ignore_etag off;` <br/>
+**Contexts:** http, server
+
+This directive can be used to disable the ETag consistency check of sliced files. It should only be used as a workaround in case the origin may generate different ETag values for the same file.
```
burtonquantil, line 614: I suggest replacing "in case" with "if"

docs/edge-logic/supported-directives.md
```diff
@@ -596,11 +596,22 @@ Assigns a value to the specified variable. No change to the public version. In p
...
-For this directive to work correctly, the origin should support HTTP status 206 for ranged requests.
+Sets the size of the slices when fetching large files from the origin. The valid values are 0, which disables slicing, OR an [nginx size](http://nginx.org/en/docs/syntax.html) that is no less than `512k` and up to `512m`. The origin has to support ranged request and respond with status code 206. When caching is desired, the statement `proxy_cache_valid 206 ...` should be used to enable caching of the ranged responses. We made the following changes to this directive on top of the open-source version:
+* We disallowed this directive in any "location" block. As a result, the entire website must use the same slice size. This is to avoid potential problems when a request needs to be processed in multiple locations.
```
burtonquantil, line 603: It is not clear  what problems are caused if "a request needs to be processed in multiple locations."

docs/edge-logic/supported-directives.md
marcfishelquantil: 1. The words "range requested" is better English than "ranged requested." Some other suggestions:
2. Instead of "...that is no less than `512k` and up to `512m`." how about one of the following alternatives:

- ...that is greater than `511k` and does not exceed `512m`.

- ...that is between `512k` and `512m` (inclusive).

docs/edge-logic/supported-directives.md
marcfishelquantil: 1. In the following sentence, "CDN360 requires all cached slices carry the same ETag value" please add the word "to" between "slices" and "carry" so it reads "CDN360 requires all cached slices to carry the same ETag value".
2. I would eliminate the word "please."

docs/edge-logic/supported-directives.md
marcfishelquantil: I agree with Burton and would suggest the following rewrite of the sentence: "Use it only as a workaround if the origin generates different ETag values for the same file."

docs/edge-logic/supported-directives.md
marcfishelquantil: Instead of " As a result, the entire website must use the same slice size. This is to avoid potential problems when a request needs to be processed in multiple locations." how about the following: "Therefore, to avoid potential problems when a request must be processed in multiple locations, the entire website must use the same slice size."

docs/edge-logic/supported-directives.md
nibinqtl: Updated. BTW, the ETA for proxy_cache_valid is for the change.

nibinqtl: @marcfishelquantil I'd like to keep the passive form to emphasize the statement.

docs/edge-logic/supported-directives.md
nibinqtl: Updated. But decided to keep "please" to be nice.

docs/edge-logic/supported-directives.md
nibinqtl: Updated.

docs/edge-logic/supported-directives.md
nibinqtl: Updated. Hopefully more clear now.

docs/edge-logic/supported-directives.md
```diff
@@ -599,9 +599,9 @@ Assigns a value to the specified variable. No change to the public version. In p
...
-* CDN360 requires all cached slices carry the same ETag value to ensure the content is consistent. When a slice fetched from the origin has a value that is different from the cached ones, any in-progress transfers to clients are terminated and all the cached slices are purged immediately. So please make sure the ETag of a file on origin does not change unless the file content really changes. This behavior can be disabled by `slice_ignore_etag on;`.
+Sets the size of the slices when fetching large files from the origin. The valid values are 0, which disables slicing, OR an [nginx size](http://nginx.org/en/docs/syntax.html) that is between `512k` and `512m`, inclusive. The origin has to support range requests and respond with status code 206. If caching is desired, the statement `proxy_cache_valid 206 ...` should be used to enable caching of the partial responses. We made the following changes to this directive on top of the open-source version:
+* We disallowed this directive in any "location" block to ensure the entire domain to have the same slice size. This is to avoid potential problems when a request needs to be processed in multiple locations with different slice sizes.
```
burtonquantil, line 603: "domain to have the same slice size" should be "domain has the same slice size."

nibinqtl: Updated

PR #93, opened 2020-09-02
Edits for Release21Aug2020-NGCDN1.1-SC88-1312

marcfishelquantil:
These content edits reflect Release21Aug2020-NGCDN1.1-SC88-1312. In addition, please see the upload following this one called "revised weight write-up for origins," which contains an updated write-up for the "Weight" button in the Origins dialog box.

I have captured low-res screenshots of the GUI screens that have changed. However, they are not included in this upload. Instead, I will work with Eleazar to capture high-res screenshots and upload them to GitHub when Eleazar returns from vacation next week.


docs/edge-logic/supported-directives.md
```diff
@@ -84,22 +84,22 @@ Sets the request variable to the given value after the authorization request com
 
 Stops processing the current set of ngx_http_rewrite_module directives. No change to the public version. 
 
-### `custom_log_field`
+### [`custom_log_field`](https://docs.cdnetworks.com/cdn/docs/edge-logic/supported-directives#custom_log_field)
 <span class="badge dark">advanced</span> <span class="badge primary">CDN360 Proprietary</span> <span class="badge yellow">ETA: Aug 2020</span>
```
burtonquantil, line 88: Your file is obsolete. You still have an ETA for custom_log_field even though it was removed in production.
https://docs.cdnetworks.com/cdn/docs/edge-logic/supported-directives#custom_log_field
Please reconcile the differences. I suspect someone else already checked in some changes

docs/portal/certificates/deleting-certificate.md
```diff
@@ -0,0 +1,5 @@
+# Deleting a Certificate
+
+1. In the left pane, click **Certificates**.
+2. Below the **Actions** column on the Certificates page, click the vertical ellipsis of the certificate you want to delete, and then select **Delete**.
```
burtonquantil, line 4: Why do you use "Below the **Actions** column" ?   The ellipsis is **IN** the Actions column.

burtonquantil (changes requested):
Please check my comments.

docs/portal/certificates/deleting-certificate.md
marcfishelquantil: It's semantics. I used "below" because it is more specific than "in." In fact, the word "below" includes the meaning of "in" and adds a location to the meaning of "in" that instructs where the ellipsis button is in the column.

burtonquantil: I think your interpretation is wrong.  Consider a spreadsheet with a column A.
If you are filling out a field at row 2 within column A, that's how you'd refer to it.  
I have never heard of anybody saying fill out row 2 **below**  column A.
That's why I'm confused.

marcfishelquantil: Either way, I do not think users will be confused. That said, we can use "in" instead of "below" if you deem it a better term to use. In fact, the word "column" implies above and below, so we can certainly use "in." I will make this change Friday and appreciate your attention to detail.

burtonquantil: Let's use "in" since you are simply referring to "column".
Here's a third party example.
https://support.google.com/docs/answer/3256570?hl=en  is an example of text written by Google about Google sheets.
They refer to a cell **in** a column.

If you had instead referred to the "Actions column heading", then something like 
"Click the ellipsis under the Actions column heading" or "Click the ellipsis below the Actions column heading." would be OK, but they're a bit verbose.     I believe "... in the Actions column" is a simpler and easier to translate if we ever needed to.

marcfishelquantil: Great. I will make this fix today.

burtonquantil (approved):
OK to merge.

PR #98, opened 2020-09-17
Marc's edit and png width changes

marcfishelquantil:
This push contains revised width sizes for png figures that appeared too small, along with various edits.


snikitin-qtl (approved):
looks good to check on dev

PR #102, opened 2020-10-08
Marc dev

marcfishelquantil:
Updated the CDN360 online help with an improved graphic for the Navigating the CDN360 User Interface topic.


burtonquantil (approved):
OK to merge.
Is the old image still used anywhere? If not, you should delete that file.

PR #104, opened 2020-10-16
Add Bucket Authentication for Private S3 buckets

snikitin-qtl:


nibinqtl (approved):
Let see this on docdev

PR #110, opened 2020-10-27
Marc dev

marcfishelquantil:
In the topic creating-property.md, I capitalized the first letters in "Edge Logic" to be consistent with the way we handle this term in other topics.


burtonquantil (approved):
OK to merge.

PR #112, opened 2020-11-03
Updated cross-reference

marcfishelquantil:
I updated a cross-reference that previously did not work on doc-dev. The update should allow the Xref to work properly.


burtonquantil (changes requested):
Your previous link didn't work on docdev.quantil.com because it is a relative link. I made the same mistake in the OpenAPI spec.

http://docdev.quantil.com/cdn/docs/edge-logic/supported-directives#origin_fast_route is an accessible URL.
Why do you think your change will go to the above?
The path in the working URL is **/cdn/docs/edge-logic/supported-directives#origin_fast_route**

burtonquantil (approved):

marcfishelquantil (commented):
I  updated the two "fast route feature" URLs to ones that work.

PR #113, opened 2020-11-03
Marc dev

marcfishelquantil:
I updated two cross-references that did not work properly on doc-dev. Both Xrefs are ""fast route feature" and can be found in the Fast Route Traffic Volume and Fast Route Traffic Requests rows in the Supported Report Types table.


burtonquantil (changes requested):
Do not hardcode the entire URL. How will that work when it gets pushed out to https://docs.cdnetworks.com?  
The production docs should not refer to docdev.quantil.com

marcfishelquantil:
@burtonquantil Xrefs have been modified to hrefs per your request.

burtonquantil (approved):
ok to merge.

PR #117, opened 2020-11-09
Release-Nov/20/2020

nibinqtl:
This should be the changes to be released on Nov 20, 2020


docs/edge-logic/supported-directives.md
```diff
@@ -352,6 +362,7 @@ For example:
 ```nginx
 origin_set_header X-Forwarded-For $client_real_ip;
 ```
+One thing to notice is that if you want to use this directive to set the `Host` header to origin, you need to make sure the "origins.hostHeader" field of [the property JSON](/cdn/apidocs#operation/createPropertyVersion) is left empty. Otherwise you will get validation error.
```
burtonquantil, line 365: You can add a comma after "Otherwise"

burtonquantil (approved):
OK to merge.

PR #119, opened 2020-11-11
Recent changes to be released

nibinqtl:


docs/edge-logic/supported-directives.md
```diff
@@ -176,7 +176,7 @@ Enables or disables adding or modifying the “Expires” and “Cache-Control
 **Default**: `gzip_types text/plain text/css text/xml text/javascript application/x-javascript application/javascript application/xml;` <br/>
 **Context**:  http, server, location
 
-CDN360 has gzip always on, and applies it to the default MIME types above. In addition, compression is only activated when the response body size is greater than 1000 bytes. The default behavior should work well for the vast majority of users. This directive can be used to enable compression on other types. The search and match are case-insensitive. We are working on supporting wildcards like `text/*` and `*javascript` with an ETA of Nov. 2020. 
+CDN360 has gzip always on, and applies it to the default MIME types above. In addition, compression is only activated when the response body size is greater than 1000 bytes. The default behavior should work well for the vast majority of users. This directive can be used to enable compression on other types. The search and match are case-insensitive. We improved the public version to support (up to 20) wildcards like `text/*` and `*javascript`.
```
burtonquantil, line 179: Remove the parentheses around  "up to 20".

docs/edge-logic/supported-directives.md
```diff
@@ -188,8 +188,21 @@ Control the server behavior based on the specified condition. Make sure you full
 ```nginx
 if ($http_x = 1 && $http_y != 2abc && $http_z) { ... }
 ```
-We support up to 9 sub-conditions. If one sub-condition is evaluated false, the subsequent ones will not be evaluated.
+*  Support the `||` operator, which performs logical OR of two sub-conditions. For example:
+```nginx
+if ($http_x = 1 || $http_y != 2abc && $http_z) { ... }
+```
+Please notice that when used together, `&&` has higher precedence than `||` and using parenthesis to group sub-conditions is not supported.
```
burtonquantil, line 195: I think "parenthesis" should be "parentheses".  @marcfishelquantil

docs/edge-logic/supported-directives.md
```diff
@@ -188,8 +188,21 @@ Control the server behavior based on the specified condition. Make sure you full
...
+We support up to 9 sub-conditions and the evaluation logic automatically skips the ones that are not affecting the final result.
 *  Support of string prefix check. The condition `$s1 ^ $s2` returns `true` if `$s1` begins with `$s2`. `$s1 !^ $s2` does the opposite.
+*  Support of integer value comparison with `<`, `<=`, `>`, `>=`. The user has to make sure both operands are valid integers, otherwise the result will be `false`. A valid integer can be either decimal, or hexidecimal with leading '0x'.
```
burtonquantil, line 198: In most of this document, you refer to "you" whereas here you switch to the third person with "The user".
It ought to be consistent.
I recommend:
"Make sure both operands are valid integers. Otherwise, the result will be `false`."

docs/edge-logic/supported-directives.md
```diff
@@ -456,7 +469,7 @@ Determines in which cases a stale cached response can be used during communicati
 **Default**:	— <br/>
 **Contexts:** http, server, location
 
-Sets caching time for different response codes. We are modifying the open-source version to support setting `time` with a variable. (ETA: Nov 2020). The specified time is only applied to responses without caching instructions from the origin. A value of 0 makes those contents not cached. If you can identify dynamic/non-cacheable contents based on request, you should use `proxy_cache_bypass` and `proxy_no_cache` to bypass caching.
+Sets caching time for different response codes. We enhanced the open-source version to support setting `time` with a variable. The specified time is only applied to responses without caching instructions from the origin. A value of 0 makes those contents not cached. If you can identify dynamic/non-cacheable contents based on request, you should use `proxy_cache_bypass` and `proxy_no_cache` to bypass caching.
```
burtonquantil, line 472: @marcfishelquantil   "only applied to"  or "applied only to"?

docs/edge-logic/supported-directives.md
```diff
@@ -582,9 +595,7 @@ Enables the specified protocols for requests to a proxied HTTPS server. No chang
...
-
-The parameter value can be an integer in [0,65536]. 0 turns off the real time logging. Variable is supported. By default, the sample rate is set for the entire site by the `realTimeLog` field of the property JSON. Use this directive to change the sample rate or turn off real-time logging selectively for some locations.
+Real-Time Log is a powerful feature of CDN360 that when enabled, sends a log entry for each request in real-time to an ingestion server of your choosing. The format of the log entry is highly configurable. To avoid overwhelming the ingestion server, the user can set a sample rate `N` which means one log entry for every N requests. N can be any integer in [0, 65536] and 0 turns off the feature. A default sample rate is specified in the [realTimeLog field of a property](</cdn/apidocs#operation/createProperty>), which is applied to all requests. This directive can be used to change the sample rate or turn off the logging selectively for some locations.
```
burtonquantil, line 598: Modify " To avoid overwhelming the ingestion server, the user can set a sample rate "  to 
"To avoid overwhelming the ingestion server, set a sample rate"

burtonquantil (changes requested):
Only a few minor suggested improvements.

docs/edge-logic/supported-directives.md
```diff
@@ -46,7 +46,7 @@ The policy parameter also supports variables; the value must be one of the three
 
 If needed, use [proxy_hide_header](#proxy_hide_header) to remove the "Cache-Control" or "Link" headers from the origin.
 
-2. The new parameter ```qtl_if(condition)``` has been introduced to allow adding the header based on some condition. If the condition is true, the ```add_header``` directive adds the header and the value to the downstream response based on the policy. The ```qtl_if``` parameter should always be at the end of the directive configuration. A condition may be any of the following,
+2. The new parameter ```if(condition)``` has been introduced to allow adding the header based on some condition. If the condition is true, the ```add_header``` directive adds the header and the value to the downstream response based on the policy. The ```if``` parameter should always be at the end of the directive configuration. A condition may be any of the following,
```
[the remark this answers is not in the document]
marcfishelquantil: @nibinqtl @burtonquantil I agree with Burton. No other comments.

docs/edge-logic/supported-directives.md
marcfishelquantil: @nibinqtl @burtonquantil Please change "...compression is only activated when the..." to "...compression is activated only when the..."
Please delete "vast."
Please change "We are working on supporting wildcards like `text/*` and `*javascript` with an ETA of Nov. 2020." to "Support for wildcards such as `text/*` and `*javascript` will be provided in November 2020." Caveat: I would advise not specifying a date in case the date slips. In this case, we could say the following: "Support for wildcards such as `text/*` and `*javascript` will be provided in the future."

docs/edge-logic/supported-directives.md
marcfishelquantil: @nibinqtl @burtonquantil Correct. Please use "parentheses" instead of "parenthesis."

docs/edge-logic/supported-directives.md
marcfishelquantil: @nibinqtl @burtonquantil I agree with Burton. Or simply remove "The user has to" and start the sentence with "Make sure both operands..." 
Please add a semicolon after "..valid integers" and add a comma after "otherwise" as follows: "...valid integers; otherwise. the result..."
Please change "hexidecimal" to "hexadecimal."
The comma after "decimal" is not necessary. I believe you added it to show that the leading 0x belongs to hexadecimal and not decimal. One way around this is the following: "A valid integer can be either a decimal value or hexadecimal value that has a leading '0x'."

docs/edge-logic/supported-directives.md
marcfishelquantil: @nibinqtl @burtonquantil Please use "applied only."
Another suggestion. Instead of "A value of 0 makes those contents not cached. If you can identify dynamic/non-cacheable contents based on request, you should use `proxy_cache_bypass` and `proxy_no_cache` to bypass caching." how about the following: "A value of 0 makes the contents not cached. If you can identify dynamic/non-cacheable contents based on request, use `proxy_cache_bypass` and `proxy_no_cache` to bypass caching."

docs/edge-logic/supported-directives.md
marcfishelquantil: @nibinqtl @burtonquantil Consider changing "of your choosing" to "of your choice." 
Consider changing the preposition "in" to "of" as follows: "The parameter value can be an integer of [0,65536]."

docs/edge-logic/supported-directives.md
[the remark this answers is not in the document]
nibinqtl: updated

docs/edge-logic/supported-directives.md
nibinqtl: updated

docs/edge-logic/supported-directives.md
nibinqtl: updated

docs/edge-logic/supported-directives.md
nibinqtl: updated

docs/edge-logic/supported-directives.md
nibinqtl: updated

docs/edge-logic/supported-directives.md
nibinqtl: This directive has been removed, due to NCAS-2434

burtonquantil (approved):
OK to merge.

PR #120, opened 2020-11-11
NCAS-2417 #comment add sprint 93 API updates to the openAPI spec

burtonquantil:


nibinqtl (approved):
Good enough to test in docdev

PR #125, opened 2020-11-24
Burton ncas2441

burtonquantil:


apispec/openapi.json
```diff
@@ -457,6 +457,15 @@
...
                           "default": "false"
+                        },
+                        "cacheKeyUri": {
+                          "type": "string",
+                          "enum": [
+                            "preRewrite",
+                            "postRewrite"
+                          ],
+                          "default": "preRewrite",
+                          "description": "This setting controls how the URI of the incoming request is incorporated into the cache key. The default setting, \"preRewrite\", puts the URI into the cache key. If you have any \"rewrite\" directive that modifies the URI, then you may wish to use the setting \"postRewrite\". This puts the modified URI into the cache key instead which can result in a higher cache hit ratio."
```
nibinqtl, line 468: This setting makes a difference only when there is "rewrite" directive in the edge logic. It is more accurate to say: "When \"rewrite\" directive is used in the edge logic to modify the request URI, this settings controls whether the value before or after the modification is incorporated into the cache key. The default setting, \"preRewrite\", puts the URI before modification into the cache key. If the \"rewrite\" directive converts multiple different URIs into the same value, you may wish to use the setting \"postRewrite\" to achieve a higher cache hit ratio."

nibinqtl (changes requested):

apispec/openapi.json
burtonquantil: @nibinqtl  I just generated the openapi.json with the changes you approved in the other pull request. Thanks.

nibinqtl (approved):
Looks good now.

PR #126, opened 2020-11-25
One more location for explaining Disable Certificate Renew Automation

marcfishelquantil:
I revised one more location where the Disable Certificate Renew Automation option needed to be explained.


burtonquantil (approved):
OK to merge.

PR #128, opened 2020-12-07
NCAS-2451 #comment updated API spec for sprint 95's changes including…

burtonquantil:
… extraServicePorts and hasConfig with inverted search


nibinqtl (approved):
Looks good to me.

PR #130, opened 2020-12-15
Removed SNI and added possible fix to standby queue cross-references in 2 deployment topics

marcfishelquantil:
This fix addresses Jira ticket NCP-2124.


burtonquantil (approved):
The removal of mention of the SNI option is OK, but I'm not sure how 
/docs/portal/certificates/managing-standby-tasks.md will be found when 
the managing-standby-tasks.md is not present under a certificates directory .  @nixoz

PR #132, opened 2020-12-17
Shortened Name of the Standby Queue Topic

marcfishelquantil:
I shortened the name of the standby topic from three words separated by hyphens to two words separated by a single hyphen. The purpose is to see whether the links from the property deployment and certificate deployment topics to the standby queue topic work on doc-dev.


burtonquantil (approved):
OK to merge

PR #134, opened 2020-12-29
Addresses Jira ticket NCP-2195

marcfishelquantil:
I added more directives than specified in the Jira ticket based on the screen shot Julia provided. Please confirm this is okay, and please let me know about any wording edits you would like to see.


nibinqtl (commented):
@marcfishelquantil , this is not the right way to introduce directives. I already added them in the "supported directives" page.

nibinqtl:
Dismiss this PR.

PR #137, opened 2020-12-31
removed reference to clicking X in date range search field

marcfishelquantil:


docs/portal/tasks/deployments.md
```diff
@@ -8,7 +8,7 @@ When a property and its associated certificates are deployed, CDN360 sends the c
 
 <p align="center"><img src="/docs/resources/images/tasks/tasks-deployment.png" alt="Deployment Tasks" width="900"></p>
  
-2. To filter results by time range, click the icon circled in the following figure, select or enter a start date and end date, and then click the icon to the right of the **End date** field. To remove the filter, click the circled icon below, and then click the **X** icon to the right of the **End date** field.
+2. To filter results by time range, click the icon circled in the following figure, select or enter a start date and end date, and then click the icon to the right of the **End date** field.
```
burtonquantil, line 11: I still don't understand the comment about clicking the icon to the right of the **End  date** field.
In Release25Dec2020's UI, the filtering  occurs without me having to click any icon. I just pick the dates and it does it.

burtonquantil (changes requested):
check the comment I made.

docs/portal/tasks/deployments.md
nibinqtl: I think @marcfishelquantil is talking about this icon:
image
But I feel we really do not need to mention this kind of details. The UI should be straightforward enough.
If we mention too much details, it also makes the document too difficult to maintain.
Because every time to make a small changes to the UI, we also need to change the document.
The document should only need to explain concepts or work flow that are unique in our product.
For example, in this case, we should only mention that by clicking the filter icon, the user will be able to specify a start and an end date to filter the results.

burtonquantil: @nibinqtl   I agree and spoke to Marc about simplifying this yesterday. I gave him some examples avoiding references to **End Date** field and icons.

marcfishelquantil: @nibinqtl I will fix this and upload a new help deliverable.

marcfishelquantil:
Please review latest upload called "Updated directives and topics describing filters."

burtonquantil (approved):
OK to merge

PR #138, opened 2021-01-13
Merge the latest changes to production

nibinqtl:


nibinqtl:
@burtonquantil @marcfishelquantil , please take a look at my changes to faq.md

docs/edge-logic/faq.md
```diff
@@ -40,8 +40,10 @@ set $cache_misc "ae=$http_accept_encoding";
 ```
 If you want to send a `Vary` header to the clients to make sure they cache different variations properly, use the [`add_header`](</docs/edge-logic/supported-directives.md#add_header>) directive. If you have to pass the `Vary` header from the origin to the client, use the following configuration to "undo" the default removal of the header:
 ```nginx
-origin_header_modify Vary "" policy=preserve; # preserve the Vary header from origin
-proxy_ignore_headers Vary; # ignore the Vary header, just pass it to the client
+# preserve the Vary header from origin
+origin_header_modify Vary "" policy=preserve;
+# ignore the Vary header, just pass it to the client
```
marcfishelquantil, line 45: How about removing the word "just" in line 45?

docs/edge-logic/faq.md
```diff
@@ -61,4 +63,22 @@ If you have one or more domains with ICP Beian and want them to be accelerated i
 
 ### How to support websocket?
 
-Use the directive [`enable_websocket`](</docs/edge-logic/supported-directives.md#enable_websocket>) in the location where websocket is needed. Make sure the client uses HTTP/1.1 (not HTTP/2) to connect. This directive also sets the read and send timeouts to 21s by default. They can be changed using the `origin_read_timeout` or `origin_send_timeout` directives.
+Use the directive [`enable_websocket`](</docs/edge-logic/supported-directives.md#enable_websocket>) in the location where websocket is needed. Make sure the client uses HTTP/1.1 (not HTTP/2) to connect. This directive also sets the read and send timeouts to 21s by default. They can be changed using the [`origin_read_timeout`](</docs/edge-logic/supported-directives.md#origin_read_timeout>) or [`origin_send_timeout`](</docs/edge-logic/supported-directives.md#origin_send_timeout>) directives.
```
marcfishelquantil, line 66: Please add the word "the" after "where" so it reads as follows: "in the location where the websocket is needed."

docs/edge-logic/faq.md
```diff
@@ -61,4 +63,22 @@ If you have one or more domains with ICP Beian and want them to be accelerated i
...
+
+If you only have the origin server at a central data center or cloud, the performance can be quite poor if the visitor is far away from or does not have good connection to the data center. How can we accelerate those contents through CDN360? Here are a few things you can do:
+* **Simply using CDN360 can shave off a few seconds**
+
+If you use CDN360, your clients will connect to our PoP that is closest to them. The round trip time (RTT) can be a few hundred milliseconds smaller compared to connecting to the origin server directly. The TCP and TLS handshakes usually take 3-4 RTTs which can easily be a second faster through CDN. The CDN360 servers by default maintain persistent connections to the origin. You can increase the [keep-alive timeout](/docs/portal/edge-configurations/managing-origins) to up to 10 minutes in the origin configurations. In the meantime, if you know the content is not cacheable at all, you should use [`proxy_no_cache 1;`](</docs/edge-logic/supported-directives.md#proxy_no_cache>) and [`proxy_cache_bypass 1;`](</docs/edge-logic/supported-directives.md#proxy_no_cache>) to completely skip caching to minimize latency.
```
marcfishelquantil, line 78: How about rewording this to the following:
If you use CDN360, your clients will connect to the CDN360 PoP that is closest to them. The round trip time (RTT) can be a few hundred milliseconds faster compared to connecting to the origin server directly. The TCP and TLS handshakes usually take 3-4 RTTs, which can be a second faster through CDN. By default, the CDN360 servers maintain persistent connections to the origin. You can increase the [keep-alive timeout](/docs/portal/edge-configurations/managing-origins) by up to 10 minutes in the origin configurations. In the meantime, if you know the content is not cacheable at all, use...

docs/edge-logic/faq.md
```diff
@@ -61,4 +63,22 @@ If you have one or more domains with ICP Beian and want them to be accelerated i
...
+* **Cache the response!**
+
+In many cases, "dynamic" does not mean the content is not cacheable at all. For example, the end users will not feel the difference if you cache the NBA score for 1 second. If you have 10 requests per second to get the score, you could save 90% of the bandwidth and processing power. If the response depends on some query paremeter or request header values, you need to make sure those variables are [included in the cache key](#how-do-you-include-query-parameters-andor-request-headers-in-the-cache-key).
```
marcfishelquantil, line 81: Please consider the following rewording. Please note that "paremeter" is misspelled.
In many cases, "dynamic" does not mean the content is not cacheable. For example, end users will not experience the difference if you cache the NBA score for 1 second. If you have 10 requests per second to fetch the score, you can save 90% of the bandwidth and processing power. If the response depends on a query parameter or request header values, make sure those variables are [included in the cache key](#how-do-you-include-query-parameters-andor-request-headers-in-the-cache-key).

docs/edge-logic/faq.md
```diff
@@ -61,4 +63,22 @@ If you have one or more domains with ICP Beian and want them to be accelerated i
...
+* **Enable Fast Route to origin**
+
+CDN360 has a directive [`origin_fast_route`](</docs/edge-logic/supported-directives.md#origin_fast_route>) to enable a "Fast Route" to access the origin. This is one of the most powerful features of CDN360. We created this based on our award-wining [High-speed Data Transmission](https://www.cdnetworks.com/enterprise-applications/high-speed-data-transmission/) (HDT) technology. It ensures our servers will have the best possible channel to reach your origin even under the most difficult situations. This directive can also be used for highly cacheable contents if the origin is hard to reach from certain networks, or when the cache miss performance is critical to you. However, the traffic served through the ""Fast Route" may be charged with a high price due to the higher cost associated with it. Please contact our support team if you want to give it a try.
```
marcfishelquantil, line 84: How about the following:
CDN360 has a directive [`origin_fast_route`](</docs/edge-logic/supported-directives.md#origin_fast_route>) that enables a "Fast Route" to access the origin. This powerful feature is based on CDN360's award-wining [High-speed Data Transmission](https://www.cdnetworks.com/enterprise-applications/high-speed-data-transmission/) (HDT) technology. This technology ensures that our servers always have the best possible channel to reach your origin, even under challenging situations. This directive can also be used for highly cacheable contents if the origin is hard to reach from certain networks or when the cache-miss performance is critical to you. However, the traffic served through the ""Fast Route" may be charged with a high price due to the higher cost associated with it. To try out this feature, contact the CDN360 support team.

marcfishelquantil:
@nibinqtl Bin, please see my feedback and let me know whether you have questions. Thank you.

docs/edge-logic/faq.md
```diff
@@ -61,4 +63,22 @@ If you have one or more domains with ICP Beian and want them to be accelerated i
 
 ### How to support websocket?
```
burtonquantil, line 64: In  http://docdev.quantil.com/cdn/docs/edge-logic/supported-directives#enable_websocket you refer to "the WebSocket protocol".   References only refer to "WebSocket" not "websocket"  https://en.wikipedia.org/wiki/WebSocket
Does this reference need to be updated?

docs/edge-logic/faq.md
burtonquantil: should that actually be "the WebSocket protocol is needed" ?

docs/edge-logic/faq.md
```diff
@@ -61,4 +63,22 @@ If you have one or more domains with ICP Beian and want them to be accelerated i
...
 
-Use the directive [`enable_websocket`](</docs/edge-logic/supported-directives.md#enable_websocket>) in the location where websocket is needed. Make sure the client uses HTTP/1.1 (not HTTP/2) to connect. This directive also sets the read and send timeouts to 21s by default. They can be changed using the `origin_read_timeout` or `origin_send_timeout` directives.
+Use the directive [`enable_websocket`](</docs/edge-logic/supported-directives.md#enable_websocket>) in the location where websocket is needed. Make sure the client uses HTTP/1.1 (not HTTP/2) to connect. This directive also sets the read and send timeouts to 21s by default. They can be changed using the [`origin_read_timeout`](</docs/edge-logic/supported-directives.md#origin_read_timeout>) or [`origin_send_timeout`](</docs/edge-logic/supported-directives.md#origin_send_timeout>) directives.
+
+### What about dynamic content?
+
+Dynamic contents are usually generated on-the-fly for each request and are different across different visitors. Some examples are:
```
burtonquantil, line 70: "across" doesn't sound like the right preposition to use.  @marcfishelquantil  I suggest changing "across different visitors to "for different visitors".

docs/edge-logic/faq.md
```diff
@@ -61,4 +63,22 @@ If you have one or more domains with ICP Beian and want them to be accelerated i
...
+* Realtime stock prices, sports game scores
+* Search results based on keywords entered by the visitor
+* API calls that have lots of query parameters
+
+If you only have the origin server at a central data center or cloud, the performance can be quite poor if the visitor is far away from or does not have good connection to the data center. How can we accelerate those contents through CDN360? Here are a few things you can do:
```
burtonquantil, line 75: I haven't heard people say "at cloud" or "at the cloud".   I suggest "in a central data center or in the cloud" instead of  "at a central data center or cloud".

Also, insert the word "a" in front of "good"  , i.e.,  " a good connection"

docs/edge-logic/faq.md
burtonquantil: A further rewrite of the second sentence could condense it to " The round trip time (RTT) can be a few hundred milliseconds faster than directly connecting to the origin server."  
I think the word "the" should be inserted before CDN.  "... faster through the CDN".

docs/edge-logic/faq.md
burtonquantil: what is "the NBA score"?  Are you referring to the score of a professional NBA game?   I think "an NBA score" might be better. Not every reader is an NBA fan, so they may not recognize "NBA".   It may be better to say "cache the score of a basketball game".

docs/edge-logic/faq.md
burtonquantil: Please change "award-wining" to "award-winning".  
Also change "charged with a high price" to "charged a higher price".    I haven't heard anyone use the word "with" like this.

docs/edge-logic/supported-directives.md
```diff
@@ -196,15 +196,15 @@ Enables or disables adding or modifying the “Expires” and “Cache-Control
 **Default**: `gzip_types text/plain text/css text/xml text/javascript application/x-javascript application/javascript application/xml;` <br/>
 **Context**:  http, server, location
 
-CDN360 has gzip always on, and applies it to the default MIME types above. In addition, compression is activated only when the response body size is greater than 1000 bytes. The default behavior should work well for the majority of users. This directive can be used to enable compression on other types. The search and match are case-insensitive. We improved the public version to support up to 20 wildcards like `text/*` and `*javascript`.
-
+CDN360 has gzip always on, and applies it to the default MIME types above. In addition, compression is activated only when the response body size is greater than 1000 bytes. The default behavior should work well for most users. This directive can be used to enable compression on other types. The search and match are case-insensitive. We improved the public version to support up to 20 wildcards like `text/*` and `*javascript`.
```
burtonquantil, line 199: I sounds better to say "CDN360 always uses gzip and applies it to the default MIME types above."

burtonquantil (changes requested):
I made a few recommendations.

docs/edge-logic/faq.md
nibinqtl: We want to emphasize the fact that nothing else is done except passing it down.

docs/edge-logic/faq.md
nibinqtl: Updated

docs/edge-logic/faq.md
nibinqtl: Updated

docs/edge-logic/faq.md
nibinqtl: Updated.

docs/edge-logic/faq.md
nibinqtl: Updated

docs/edge-logic/faq.md
nibinqtl: Updated

docs/edge-logic/faq.md
nibinqtl: Updated

docs/edge-logic/faq.md
nibinqtl: Updated

docs/edge-logic/supported-directives.md
nibinqtl: Updated.

nibinqtl:
I believe I addressed all suggestions. Please review again. @burtonquantil @marcfishelquantil

burtonquantil (dismissed):
OK to merge

nibinqtl:
@burtonquantil , I just merged your #139 and now this PR needs approval again.

burtonquantil (approved):
Thanks @nibinqtl  . OK to merge.

PR #157, opened 2021-02-23
addresses NCP-2277

marcfishelquantil:
This upload addresses Julia Korsun's comment that protocol was mentioned in the field descriptions after the first Report screenshot.


burtonquantil (approved):
OK to merge.

PR #163, opened 2021-03-31
To be released on 4/7

nibinqtl:


nibinqtl:
@marcfishelquantil @burtonquantil , please help review the new article about secure content delivery

nibinqtl:
Thanks @marcfishelquantil . I merged your edits and made some additional changes. I may make some more changes before the release. Will need you to review again.

docs/recipes/secure-delivery.md
```diff
@@ -0,0 +1,117 @@
+## Secure Content Delivery with CDN360
+
+Security and protection of privacy have raised increasing concerns over the last few years — and for good reason. Hardly a day goes by when the headlines aren't filled with another major security breach. With these concerns in mind, CDN360 has adopted several features to help you achieve your security goals with smooth and uninterrupted service for your clients. This article describes these features, along with a few best practices to help you optimize security.
```
burtonquantil, line 3: Simplify the first sentence.
"Security and protection of privacy are increasing concerns over the last few years"

docs/recipes/secure-delivery.md
```diff
@@ -0,0 +1,117 @@
...
+
+### Layer 4 DDoS Mitigation
+At the entry point of every CDN360 Point of Presence (PoP) is a high-performance Layer 4 distributed denial-of-service (DDoS) firewall. The firewall consists of a group of machines that analyze incoming traffic at line speed.
+Based on regularly updated rules, the firewall rejects suspicious packets that may endanger services and forwards only the "safe" packets to the servers located behind the firewall. This feature is enabled for all services and is transparent to customers.
```
burtonquantil, line 7: "transparent to customers"  - Since you began the article addressing the audience with "your" and "you", shouldn't this be "... transparent to you"  ?

docs/recipes/secure-delivery.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+### Access Control to the Origin
+It is always a good idea to set up some ACL rules on the origin to avoid spamming. In this case, the [`eval_func`](</docs/edge-logic/supported-directives.md#eval_func>) directive can also be used to generate the required token for accessing the origin. Here is an example of how to implement the [AWS Signature Version 2](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html):
+```nginx
+##required input variables: $awskey $awsseckey $awsbucket/$s3key
+#Step 1: construct STS
+set $awsdatev2 $time_rfc822;
+set $awssts "GET\n\n\n$awsdatev2\n/$awsbucket/$s3key";
+#Step 2: sign STS with secKey
```
burtonquantil, line 67: Minor suggestion: I think inserting an empty row between the steps improves readability, though of course, it is irrelevant to the software.

docs/recipes/secure-delivery.md
```diff
@@ -0,0 +1,117 @@
...
+</html>';
+    }
+    # continue loading the page from origin or cache
+    origin_pass my_origin;
+}
+```
+More sophisticated methods can be adopted in this way to block more advanced bots.
+
+### TLS features
+* CDN360 supports TLS certificates with both RSA and ECDSA algorithms. You can even configure 2 certificates with different algorithms in the same property and have the server pick one based on the client's capability and preference.
```
burtonquantil, line 102: I'd use "two" instead of "2" , but it's OK if you wish to leave it as is.

docs/recipes/secure-delivery.md
```diff
@@ -0,0 +1,117 @@
...
+* We highly recommend that you set the minimum TLS version to 1.2. For maximum security and performance, however, you should really take advantage of TLSv1.3.
+* CDN360 also allows you to fully configure TLS ciphers based on your security requirements. For example, you can prioritize the ECDHE and EDH key exchange algorithms to ensure "[Perfect Forward Secrecy](https://www.digicert.com/kb/ssl-support/ssl-enabling-perfect-forward-secrecy.htm)".
+* If a client request uses HTTPS, CDN360 contacts the origin with the same protocol to ensure that the entire path is encrypted. Although CDN360 supports "protocol downgrades," you should avoid using them unless absolutely necessary.
+* To avoid "man-in-the-middle" attacks or DNS hijacking attempts of your origin's hostname, enable the validation of the origin's certificate.
+* If your site supports HTTPS, a good practice is to redirect all HTTP requests to the HTTPS counterpart. You can achieve this on the [CDN360 portal](/docs/portal/edge-configurations/creating-property.md#tls-settings) with a simple drop-down list.
```
burtonquantil, line 107: I would rewrite the last sentence slightly
"You can achieve this **using** the [CDN360 portal](/docs/portal/edge-configurations/creating-property.md#tls-settings) **by selecting an option from a list**.

burtonquantil (changes requested):
I made a few small suggestions.

docs/recipes/secure-delivery.md
nibinqtl: made a small change but decided to keep the "raised".

docs/recipes/secure-delivery.md
nibinqtl: changed to "all the users"

docs/recipes/secure-delivery.md
nibinqtl: rewrote with a screenshot of the dropdown list

nibinqtl:
Hi @burtonquantil and @marcfishelquantil , please take a look at all the changes I have made after your last reviews.

docs/recipes/secure-delivery.md
marcfishelquantil: In the last sentence, I would suggest replacing "with" with "using". I would also remove "simply." Here are some suggestions to consider that include or omit "with" and "simply."

"You can achieve this on the [CDN360 portal](/docs/portal/edge-configurations/creating-property.md#tls-settings) using a drop-down list." 

"You can achieve this on the [CDN360 portal](/docs/portal/edge-configurations/creating-property.md#tls-settings) simply by using a drop-down list."

"You can easily achieve this on the [CDN360 portal](/docs/portal/edge-configurations/creating-property.md#tls-settings) by using a drop-down list."

"You can achieve this easily on the [CDN360 portal](/docs/portal/edge-configurations/creating-property.md#tls-settings) by using a drop-down list."

"Redirecting HTTP requests to HTTPS on the [CDN360 portal](/docs/portal/edge-configurations/creating-property.md#tls-settings) is as easy as selecting a the redirect option in a drop-down list."

"To redirect HTTP requests to HTTPS on the [CDN360 portal](/docs/portal/edge-configurations/creating-property.md#tls-settings), just select the redirect option from a drop-down list."

burtonquantil (approved):
OK to merge

PR #182, opened 2021-05-25
NCAS-2757 #comment added descriptions of API updates for sprint t 107…

burtonquantil:
…, Release14May2021. These have been checked by QA

### DIFF
--- a/apispec/openapi.json
+++ b/apispec/openapi.json
@@ -23682,7 +23682,7 @@
                       "edgeHostnames": {
                         "type": "boolean",
                         "default": false,
-                        "description": "Whether the API account can access the edge hostname management (/cdn/edgeHostnames) endpoints. By default, read-only API accounts can call the endpoints using the HTTP GET method while the operator API accounts do not have access."
+                        "description": "Whether the API account can access the edge hostname management (/cdn/edgeHostnames) endpoints."
                       },
                       "validations": {
                         "type": "boolean",
@@ -23718,7 +23718,7 @@
                       "apiPermissions": {
                         "type": "boolean",
                         "default": true,
-                        "description": "Whether the API account can access the API permission  (/cdn/apiPermissions) endpoints. By default, read-only API accounts can call the endpoints using the HTTP GET method while the operator API accounts do not have access."
+                        "description": "Whether the API account can access the API permission  (/cdn/apiPermissions) endpoints."
                       },
                       "systemConfigs": {
                         "type": "boolean",
--- a/apispec/openapi.json
+++ b/apispec/openapi.json
@@ -20081,7 +20081,7 @@
                   },
                   "allowProduction": {
                     "type": "boolean",
-                    "default": "false",
+                    "default": false,
                     "description": "Indicates if deployment to production servers is permitted. Deployment to staging servers is always permitted."
                   },
                   "advancedFeatures": {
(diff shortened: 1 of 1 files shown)


apispec/openapi.json
```diff
@@ -23617,7 +23682,7 @@
                       "edgeHostnames": {
                         "type": "boolean",
                         "default": "false",
-                        "description": "Whether the API account can access the edge hostname management (/cdn/edgeHostnames) endpoints."
+                        "description": "Whether the API account can access the edge hostname management (/cdn/edgeHostnames) endpoints. By default, read-only API accounts can call the endpoints using the HTTP GET method while the operator API accounts do not have access."
```
nibinqtl, line 23685: I think this is not correct. The operator API account should be able to see everything.

nibinqtl: Same problem for the fields below.

apispec/openapi.json
```diff
@@ -24496,78 +24578,658 @@
...
...e> <code> $ dig staging.qtlcdn.com <br/> ... <br /> ;; ANSWER SECTION:<br/> staging.qtlcdn.com.    59      IN      A       163.171.228.89<br/> staging.qtlcdn.com.    59      IN      A       163.171.230.47<br/> </code> </pre> You can also call the <a href=\"#operation/getStagingServers\">Query list of staging servers API</a> to get IP addresses you can use. You have several ways to set up your system to test against staging: <ol> <li>Modify your DNS server to point your deployed property's hostname to the CDN360 staging hostname <i>staging.qtlcdn.com</i>.</li> <li>Modify your HOSTS file (/etc/hosts on Linux) to hard code the property's hostname to one of the staging servers' IP addresses.</li> <li> If the cURL command is used, the --resolve option can be used to directly map the property's hostname to an IP address of a CDN360 staging server.</li> </ol> Once you have done the setup, you can access your content via the hostname. <h2>Testing a property deployed to production</h2> After you <a href=\"#operation/createEdgeHostname\">create an edge hostname</a> and set your DNS server to point the edge hostname to your property's hostname, you can browse your content under the hostname."
```
nibinqtl, line 25184: We'd better provide a link to the documentation page about staging test, instead of maintaining the details here.

nibinqtl (changes requested):
A few changes are needed.

apispec/openapi.json
burtonquantil: @nibinqtl  I'm pretty sure this is what our current implementation is doing even though it sounds odd.  Refer to https://quantil.atlassian.net/browse/NCAS-2757?focusedCommentId=202333

apispec/openapi.json
burtonquantil: That's a good idea @nibinqtl .
However, I just found another reference related to staging. We had an API returning the list of IP addresses.
https://docs.cdnetworks.com/cdn/apidocs#operation/getStagingServers
The backward incompatible change we made to the staging IP address behavior means the above instructions aren't really usable as is.  A customer could use one that is for China only without realizing it.

burtonquantil: I'm thinking of replacing that text with:

 
> We highly recommend you <a href="https://docs.cdnetworks.com/cdn/docs/portal/edge-config\
> urations/testing-property">deploy to our <i>staging</i> environment to test your configurations before deploying to <i>production</i></a>.

This gets rid of the big heading <h2>Testing a staged property</h2>  though. What do you think?

marcfishelquantil:
These look good to me. /Marc

apispec/openapi.json
burtonquantil: Lijo explained the implementation. I will correct the text here.

burtonquantil:
@nibinqtl  Please check again as I made some updates.

nibinqtl (approved):
Good enough to see it in dev

PR #186, opened 2021-06-19
Latest changes to be released by 6/25

nibinqtl:


docs/edge-logic/multiple-origins.md
```diff
@@ -15,6 +15,39 @@ location / {
...
+*   Keep in mind that only the directives in the rewrite module ([`if`](</docs/edge-logic/supported-directives.md#if>), [`set`](</docs/edge-logic/supported-directives.md#set>), [`rewrite`](</docs/edge-logic/supported-directives.md#rewrite>), [`return`](</docs/edge-logic/supported-directives.md#return>), [`break`](</docs/edge-logic/supported-directives.md#break>), [`eval_func`](</docs/edge-logic/supported-directives.md#eval_func>), etc.) are executed in the order they appear (imperatively) in the location block. They are executed in an early phase which is before most other actions defined by the declarative directives. In the edge logic editor on the portal, the imperative directives are colored in blue while the declarative ones are colored in red so you can easily tell them apart. When there are declarative directives enclosed in `if` blocks, only the ones in the **last** matching `if` block will take effect. For example, with the configuration below:
+```nginx
```
burtonquantil, line 19: I suggest "For example, with the configuration below:" be rewritten as "For example, consider the following configuration:"

docs/edge-logic/multiple-origins.md
```diff
@@ -15,6 +15,39 @@ location / {
...
+  if ($http_header_a != '') {
+    add_header has-header-a 1;
+  }
+  if ($http_header_b != '') {
+    add_header has-header-b 1;
+  }
+  ...
+}
+```
+When both `header-a` and `header-b` are present in the request, only `has-header-b` will be added to the response. This is not a bug, just how NGINX works. But it can be confusing to new users.
```
burtonquantil, line 30: I don't believe we should begin sentences with "But".
Perhaps rewrite as:
"This is not a bug, just how NGINX works, but this can be confusing to new users."

docs/edge-logic/multiple-origins.md
```diff
@@ -15,6 +15,39 @@ location / {
...
+*   Due to the issue mentioned above, you are recommended to refrain from using any declarative directive in `if` blocks, unless there is no other way around. This makes the configuration more readable, especially to people who are not NGINX expert. The example above can be rewritten in the way below to achieve the desired result:
```
burtonquantil, line 31: - Since you referred to "people" which is indicates multiple individuals, "expert" should be "experts" .
-  Suggestion: "The example above can be rewritten as follows to achieve the desired result:"

burtonquantil (changes requested):
I made several suggestions.

The apispec/openapi.json included in this pull request does not cover Release25Jun2021. I have an open bug to work on that update.

docs/edge-logic/multiple-origins.md
marcfishelquantil: Some suggestions:
- Please change "of NGINX configuration" to "of the NGINX configuration."
- Please change "For example, try to understand" to "In particular, try to understand."
- Please change "They are executed in an early phase which is before most..." to "They are executed in an early phase, which precedes most ..." Note that this phrase appears twice above.
- To have the adverb appear properly at the end of the sentence, please change "...so you can easily tell them apart" to "...so you can tell them apart easily." Note that this phrase appears twice above.
-

docs/edge-logic/multiple-origins.md
marcfishelquantil: Please reword "This is not a bug, just how NGINX works. But it can be confusing to new users." to "This how NGINX works and is not a bug. However, this behavior can be confusing to new users."

marcfishelquantil:
Marc provided comments.

nibinqtl:
@marcfishelquantil did you hit "Close" by accident?

nibinqtl:
Updated, please take another look.

burtonquantil (approved):
OK to merge.  However, please note Marc merged his portal help updates related to Release25Jun2021 to dev.  He describes Secret Management.  We have not deployed this release to production yet.

nibinqtl:
OK. Lets release this right after we release the code.

docs/edge-logic/multiple-origins.md
```diff
@@ -12,9 +12,42 @@ location / {
   }
 }
 ```
-<a id="ifcaution"></a>This example uses the [`if` directive in the rewrite module](http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#if) to check the value in the request header `x-origin` and define different behaviors. However, `if` is a tricky directive due to the conflict of the "[declarative](https://tylermcginnis.com/imperative-vs-declarative-programming/)" nature of NGINX configuration and the "imperative" nature of the rewrite module. If you need to use `if` in your Edge Logic, observe the following rules:
+<a id="ifcaution"></a>This example uses the [`if` directive in the rewrite module](http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#if) to check the value in the request header `x-origin` and define different behaviors. However, `if` is a tricky directive due to the conflict of the "[declarative](https://tylermcginnis.com/imperative-vs-declarative-programming/)" nature of the NGINX configuration and the "imperative" nature of the rewrite module. If you need to use `if` in your Edge Logic, observe the following rules:
```
burtonquantil, line 15: "conflict of" should be "conflict between"

burtonquantil (dismissed):
OK to merge

burtonquantil (approved):
OK to merge.   Ele made a big pull request to duplicate the help text into cn and en subdirectories in order to investigate the possibility of supporting multiple languages that way. He said he needed these to test locally with Lijo. The production docs.cdnetworks.com should not be affected by these two new directories since docs-app doesn't refer to them yet.

PR #202, opened 2021-08-17
NGDOC-97 #Uploading video

eleazarmadera:
Uploading Getting Started Video


burtonquantil (approved):
Looks good.. I don't know if we'll ever translate to Chinese, though.

PR #221, opened 2021-09-13
Updated 2 screens per NCP-2585 and NCP-2572.

marcfishelquantil:
Per NCP-2585, I updated two screens to show the new tooltip icon next to the Let's Encrypt option on the Edit Certificate page. NCP-2572 does not affect the documentation because we do not show tooltip definitions in the documentation.


burtonquantil (approved):
These changes look OK.

PR #222, opened 2021-09-14
Added info about nearChina server group

marcfishelquantil:
Burton, I updated 2 topics to cover this new server group.


burtonquantil (changes requested):
Changes look OK. Can you also edit the files under the zh directory?
These will eventually be translated to Chinese.

burtonquantil (approved):
OK to merge

PR #231, opened 2021-09-22
NGDOC-106

burtonquantil:
This consists of release notes covering Release17Sep2021 of our CDN360 portal/API which was deployed on September 21, 2021

### DIFF
--- a/en/docs/release-notes.md
+++ b/en/docs/release-notes.md
@@ -1,5 +1,21 @@
 # Release Notes
 
+## September 21, 2021
+
+### API updates
+* Disallowed update of the edgeHostname field in edge hostnames because it can cause service interruptions.
+* Disallowed update of the hasBeian field in edge hostnames because of planned changes to edge hostnames to use different suffixes for those with ICP Beian versus those without.
+
+### Portal updates
+* Fixed display of plan page for self-signup trial users.
+* Updated warning text about pending standby task if the user tries to log out before deploying.
+* Improved response to erroneous input in the report page.
+* Improved display of lengthier tooltip in the property configuration page.
+* Fixed issues supporting Russian translation.
+* Added clarification for trial users that deployment to production is enabled only after they’ve done some tests against staging.
+* Supported soft wrap in the Edge Logic editor.
+* Allowed Edge Logic hint text to be translated. Directive names are keywords that must remain in English.
+
 ## September 8, 2021
 ### API updates
 * Returned error code InvalidApiAccountId if an incorrect API account ID is specified when calling [API permissions](</apidocs#tag/API-Access-Control>) endpoints.
--- a/en/docs/release-notes.md
+++ b/en/docs/release-notes.md
@@ -3,16 +3,16 @@
 ## September 21, 2021
 
 ### API updates
-* Disallowed update of the edgeHostname field in edge hostnames because it can cause service interruptions.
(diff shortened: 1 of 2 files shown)


en/docs/release-notes.md
```diff
@@ -1,5 +1,21 @@
 # Release Notes
 
+## September 21, 2021
+
+### API updates
+* Disallowed update of the edgeHostname field in edge hostnames because it can cause service interruptions.
```
marcfishelquantil, line 6: Suggestion. Instead of "...because it can cause service interruptions." please consider "...because it can interrupt service."

en/docs/release-notes.md
```diff
@@ -1,5 +1,21 @@
...
+* Disallowed update of the hasBeian field in edge hostnames because of planned changes to edge hostnames to use different suffixes for those with ICP Beian versus those without.
+
+### Portal updates
+* Fixed display of plan page for self-signup trial users.
+* Updated warning text about pending standby task if the user tries to log out before deploying.
+* Improved response to erroneous input in the report page.
+* Improved display of lengthier tooltip in the property configuration page.
```
marcfishelquantil, line 13: "lengthier" is a comparative that compares 2 items. In this case, I would suggest replacing it with "lengthy." Also, "tooltip" should be the plural "tooltips."

en/docs/release-notes.md
```diff
@@ -1,5 +1,21 @@
...
+* Fixed issues supporting Russian translation.
+* Added clarification for trial users that deployment to production is enabled only after they’ve done some tests against staging.
```
marcfishelquantil, line 15: Please consider changing "...only after they’ve done some tests..." to "...only after they’ve performed some tests...".

zh/docs/release-notes.md
```diff
@@ -1,5 +1,21 @@
 # Release Notes
 
+## September 21, 2021
+
+### API updates
+* Disallowed update of the edgeHostname field in edge hostnames because it can cause service interruptions.
```
marcfishelquantil, line 6: Suggestion. Instead of "...because it can cause service interruptions." please consider "...because it can interrupt service."

zh/docs/release-notes.md
```diff
@@ -1,5 +1,21 @@
...
+* Disallowed update of the hasBeian field in edge hostnames because of planned changes to edge hostnames to use different suffixes for those with ICP Beian versus those without.
+
+### Portal updates
+* Fixed display of plan page for self-signup trial users.
+* Updated warning text about pending standby task if the user tries to log out before deploying.
+* Improved response to erroneous input in the report page.
+* Improved display of lengthier tooltip in the property configuration page.
```
marcfishelquantil, line 13: "lengthier" is a comparative that compares 2 items. In this case, I would suggest replacing it with "lengthy." Also, "tooltip" should be the plural "tooltips."

zh/docs/release-notes.md
```diff
@@ -1,5 +1,21 @@
...
+* Fixed issues supporting Russian translation.
+* Added clarification for trial users that deployment to production is enabled only after they’ve done some tests against staging.
```
marcfishelquantil, line 15: Please consider changing "...only after they’ve done some tests..." to "...only after they’ve performed some tests...".

marcfishelquantil (approved):

PR #249, opened 2021-10-28
changes to be release to production by 11/5

nibinqtl:


en/docs/edge-logic/supported-directives.md
```diff
@@ -159,11 +159,20 @@ This directive is very similar to the [`send_timeout`](http://nginx.org/en/docs/
 
 <span class="badge dark">advanced</span> <span class="badge primary">Proprietary</span>
 
-**Syntax:** `custom_log_field {custom log field id} {value or variable};`<br/>
+**Syntax:** `custom_log_field id value;`<br/>
 **Default:** `-`<br/>
 **Context:** server, location, if in location
 
-This directive allows you to add up to 2 customized fields into the access log. They can be referred to by the keywords "custom1" and "custom2" when you configure the format of the download log or when using our advanced traffic analysis tool. If you require this feature, contact our support team.
+This directive allows you to add up to 2 customized fields into the access log. The id can be either 1 or 2. The value can contain variables. The 2 fields can be referred to by the keywords "custom1" and "custom2" when you configure the format of the download log or when using our advanced analytical tool. If you require this feature, contact our support team.
```
burtonquantil, line 166: Minor: "The 2 fields" should be "The two fields"

en/docs/intro.md
```diff
@@ -1,10 +1,8 @@
-# Welcome to CDN Pro
+# Welcome to CDN Pro (previously known as CDN360)
 
-CDN Pro is leading the next evolution of content delivery networks (CDN). 
+CDN Pro is a platform we architected from scratch based on the latest technology and our 20+ years of practice in the content delivery industry. It is designed to give you unprecedented flexibility to deliver digital assets to consumers globally with the best possible performance and security. Under the hood, we made sure the system has a very clean structure that is easy to operate and extend in the future. These factors are important to keep the cost down and outages away.
```
burtonquantil, line 3: The sentence is valid English, but I prefer the word "experience" to "practice".  The word has different meanings. One is doing something over and over to get better at.  Yes, the new CDN is better, but I don't want to imply too much that the old system was poor. It is a real product that was effective as well.

@ronyayoung

en/docs/intro.md
```diff
@@ -1,10 +1,8 @@
...
 
-CDN Pro is designed to accelerate and deliver your applications to users worldwide, regardless of their location. It boosts content delivery, improves load times, and offers many more clear-cut advantages to deliver an unparalleled user experience.
-
-In addition to accelerating, securing, and scaling your applications, CDN Pro enables you to have granular control over your content delivery and costs. Unique self-service capabilities, along with a friendly, yet powerful user interface and application programming interface (API), provide for seamless integration into your SDLCs.
+The platform comes with a friendly yet powerful user interface that allows 100% self-service of even the most complicated delivery configurations. Its application programming interface (API) enables seamless integration of CDN into the entire development lifecycle of your applications. We also extended the concept of 100% self-service to the business side. CDN Pro has a unique price model that allows you to explore for the best trade-off between performance and cost. We truly believe this is a product that will change the industry!
```
burtonquantil, line 5: Minor:  "to explore for the best trade-off between performance and cost."
I prefer:  "to balance performance and cost."
The word "trade-off" is certainly valid. It refers to balancing of things that cannot be attainable at the same time. To me, it has more negative connotations than "balance"... like you will  get good performance from our CDN only if you pay the most. This isn't necessarily true.  
@ronyayoung

en/docs/manifest.json
```diff
@@ -2,7 +2,7 @@
     "metadata": {
       "id": "cdn",
       "title": "CDN Pro",
-      "description": "CDN Pro leads the evolution of content delivery networks (CDNs) by accelerating, securing, scaling, and delivering content to website visitors worldwide, regardless of their location.",
+      "description": "CDN Pro (a.k.a. CDN360) leads the evolution of the industry by giving the content owners unprecedented flexibility to deliver their digital assets globally with the best possible performance and security.",
```
burtonquantil, line 5: I prefer "previously known as CDN360" which was used in the intro.md to "a.k.a. CDN360" . Not everyone will recognize that abbreviation. Also, it should have been written as "aka" if you really want to use that.

burtonquantil (changes requested):
I made a few suggestions on wording. 
They are not critical.  At a minimum, replace "a.k.a." which is not a valid abbreviation or word.

nibinqtl:
Updated. Please review again. @burtonquantil

en/docs/edge-logic/supported-directives.md
```diff
@@ -159,11 +159,20 @@ This directive is very similar to the [`send_timeout`](http://nginx.org/en/docs/
 
 <span class="badge dark">advanced</span> <span class="badge primary">Proprietary</span>
 
-**Syntax:** `custom_log_field {custom log field id} {value or variable};`<br/>
+**Syntax:** `custom_log_field id value;`<br/>
 **Default:** `-`<br/>
 **Context:** server, location, if in location
 
-This directive allows you to add up to 2 customized fields into the access log. They can be referred to by the keywords "custom1" and "custom2" when you configure the format of the download log or when using our advanced traffic analysis tool. If you require this feature, contact our support team.
+This directive allows you to add up to 2 customized fields into the access log. The id can be either 1 or 2. The value can contain variables. The two fields can be referred to by the keywords "custom1" and "custom2" when you configure the format of the download log or when using our advanced analytical tool. If you require this feature, contact our support team.
```
burtonquantil, line 166: Minor: "2 customized fields" is better as "two customized fields"
"The two fields can be referred to by the keywords "custom1" and "custom2" when you configure the format of the download log or when using our advanced analytical tool"  could be written as
'Refer to the two fields using the keywords "custom1" and "custom2" when configuring the download log format or when using our advanced analytical tool.'

By the way, what exactly does "advanced analytical tool" refer to?

en/docs/intro.md
```diff
@@ -1,10 +1,8 @@
-# Welcome to CDN Pro
+# Welcome to CDN Pro (previously known as CDN360)
 
-CDN Pro is leading the next evolution of content delivery networks (CDN). 
+CDN Pro is a platform we architected from scratch based on the latest technology and our 20+ years of experience in the content delivery industry. It is designed to give you unprecedented flexibility to deliver digital assets to consumers globally with the best possible performance and security. Under the hood, we made sure the system has a very clean structure that is easy to operate and extend in the future. These factors are important to keep the cost down and outages away.
```
burtonquantil, line 3: "These factors are important to keep the cost down and outages away."
"outages away" is not normal English.
Suggestion:  "This helps reduce costs and outages."
Alternative:  combine into one sentence...
"Under the hood, we made sure the system has a very clean structure that is easy to operate and extend in the future, thereby reducing costs and outages."

Marc might have other suggestions.

en/docs/intro.md
```diff
@@ -1,10 +1,8 @@
...
 
-CDN Pro is designed to accelerate and deliver your applications to users worldwide, regardless of their location. It boosts content delivery, improves load times, and offers many more clear-cut advantages to deliver an unparalleled user experience.
-
-In addition to accelerating, securing, and scaling your applications, CDN Pro enables you to have granular control over your content delivery and costs. Unique self-service capabilities, along with a friendly, yet powerful user interface and application programming interface (API), provide for seamless integration into your SDLCs.
+The platform comes with a friendly yet powerful user interface that allows 100% self-service of even the most complicated delivery configurations. Its application programming interface (API) enables seamless integration of CDN into the entire development lifecycle of your applications. We also extended the concept of 100% self-service to the business side. CDN Pro has a unique price model that allows you to balance performance and cost. We truly believe this is a product that will change the industry!
```
burtonquantil, line 5: Instead of referring to "The platform" and "Its application programming interface", you could 
say "Our platform" and "Our application programming interface" since you do use the word "We" twice.

burtonquantil (changes requested):
made some additional suggestions.

en/docs/edge-logic/supported-directives.md
nibinqtl: updated.

en/docs/intro.md
nibinqtl: Made some changes, but decided to keep "outages away"

en/docs/intro.md
nibinqtl: Use "Our" for the first sentence.

burtonquantil (approved):
"outages away" is not English that would be written by people in the US, but we won't block the deployment for that.

PR #266, opened 2021-12-03
Release26Nov2021 changes

marcfishelquantil:
Burton, please review and provide feedback. I did not upload the equivalent Chinese topic because I want to check with Kevan first to maker sure I do not overwrite any translation he may have done. Thank you.


burtonquantil (approved):
OK to merge

PR #270, opened 2021-12-22
Revised figure reference

marcfishelquantil:
Burton, please confirm that the content management topic is bug-free.


burtonquantil (approved):
this seems to be what was in the file before. Let's give it a try.

PR #274, opened 2021-12-22
NGDOC-128 #comment added release notes for Release10Dec2021 which was…

burtonquantil:
… deployed on December 21


marcfishelquantil (approved):
All looks good. No suggestions.

PR #275, opened 2021-12-22
Updated links to point to subtopic headings

marcfishelquantil:
Burton, these fixes should point users to the children headings instead of to the line below them.


burtonquantil (approved):
OK let's give it a shot

PR #278, opened 2021-12-31
Deleted space in the scheme pattern

marcfishelquantil:
Burton, I deleted the space and found the delimiter to use to eliminate the frowning face"


burtonquantil (approved):
ok to merge

PR #290, opened 2022-01-25
Improve the break directive description according to NCAS-3160

Keven-Lin:


nibinqtl (approved):
I made some improvements.

PR #298, opened 2022-02-09
Add description about $cookie_xx variable

Keven-Lin:
Add description about $cookie_xx variable


nibinqtl (approved):
Made some changes. Now look good.

PR #301, opened 2022-02-23
Prefetch Edits and Screenshots

marcfishelquantil:
Please confirm that the text edits and screenshots are good.


burtonquantil (approved):
OK to merge.
I believe the description of regions isn't entirely accurate since we don't have servers everywhere. I will ask my team member if we need to revise the implementation and documentation.

PR #304, opened 2022-02-28
Revised region text

marcfishelquantil:
I revised the wording for Regions in the table below step 3 in the Creating a Prefetch Request topic to confirm with Burton's suggestion.


burtonquantil (approved):
OK to merge

PR #310, opened 2022-03-10
Updated path in topic

marcfishelquantil:
Attempt to fix block preventing online help from appearing in docdev.


burtonquantil (approved):
ok to merge

PR #320, opened 2022-04-06
fixed Using Properties to Configure the Edge

marcfishelquantil:
Fixed two topics to correct numbering under Using Properties to Configure the Edge > Using the Advanced Filter Icon.


burtonquantil (approved):
ok to merge

PR #337, opened 2022-05-10
revised paths in json file

marcfishelquantil:
I updated the paths in the json file. Please approve so we can see whether the error is corrected.


burtonquantil (approved):
OK to merge

burtonquantil:
@marcfishelquantil  sorry, I should have looked more carefully, the parentheses and commas you added are suspicious., especially the one injected into the array. It breaks the build.

PR #340, opened 2022-05-10
adding missing graphics to zh folder

marcfishelquantil:
This upload should fix missing graphics in the zh folder.


burtonquantil (changes requested):
Per our Skype chat, can you choose Chinese in the portal and recapture the screenshots?
That would be more appropriate for the Chinese language help pages.

burtonquantil (approved):
You can make the recapture of screenshots in Chinese another time.

PR #345, opened 2022-05-16
Dev files with Chinese GUI snapshots

marcfishelquantil:
Burton, please confirm.


burtonquantil (approved):
zh/docs/resources/images/edge-configurations/property-deployment.png also looks a little different

PR #347, opened 2022-05-16
Updated screenshots

marcfishelquantil:
Updated one English screen shot and added 2 Chinese screen shots that were mistakenly shown in English.


nibinqtl (changes requested):
@marcfishelquantil , almost all the screenshots you made have these 2 common problems:
1. The resolution is much lower than the original ones. For example, this picture: zh/docs/resources/images/edge-configurations/property-versions.png, the original one has 1940x364 pixels. Yours has only 801x170
2. The size of the text in the picture is proportionally too small. They are really hard to read.
I think you need to redo all the images.

marcfishelquantil:
@nibinqtl  Bin thank you for your suggestion. Yesterday, Eleazar tried your suggestion, among others. The results were less than satisfactory. He and I confirmed that my graphics card supports the required higher resolution needed to capture crisp screen shots, but that my monitor is the limiting factor. Thous high-res, it cannot reach the 4k resolution needed to capture acceptable screen shots. To rectify this situation, I have identified three 4k monitors and will check availability my Wednesday. I'll provide an update accordingly, as I realize I need to be able to accommodate this requirement to be as helpful as possible to the company.

zh/docs/portal/reports/generating-reports.md
```diff
@@ -55,4 +55,30 @@ The following report types are supported:
...
+After you generate a report, you can create text annotations to highlight events related to traffic going to your hostnames.
+
+1. Click the **Download Options** button at the top right of the report and select **Create Annotation**.
+2. Complete the fields in the Create an annotation dialog box.
+
+   <p align=center><img src="/docs/resources/images/reports/create_annotation.png" alt="report page" width="900"></p>
+
+| **Fields**      | **Description**                                       |
+| --------------- | ----------------------------------------------------- |
+| Annotation Type | Select the type of annotation you want to add.<ul><li> <strong>Client</strong> = an event affecting clients, such as the release of a popular game.</ul></li><ul><li><strong>Configuration</strong> = an event related to a configuration change.</ul></li><ul><li><strong>Operation</strong> = an operation performed on the CDN platform, such as updating cache servers. This annotation type is available to superusers only.</ul></li>  |
```
burtonquantil, line 71: @marcfishelquantil  please do not mention "superusers".
The customer cannot create the Operation annotation. It will not be an option for them, so it should be omitted from this section.
It can be referenced in your description of the reports page showing annotations. In that case, you can also mention something like:
"Annotations created by CDNetworks to indicate operations events may also appear in your reports."

burtonquantil (changes requested):
Please make an adjustment to the help about annotations.

zh/docs/portal/reports/generating-reports.md
marcfishelquantil: @burtonquantil Thank you for pointing this out. I revised the definition for Annotation Type accordingly.

burtonquantil (approved):
OK to merge

PR #349, opened 2022-05-20
Deleted link and resized annotation screen shot

marcfishelquantil:
In the Generating Reports topic, I deleted the bad link pointing to the "Creating Annotations" topic, which appears at the end of the "Generating Reports" section, and resized two_annotations.png.


burtonquantil (approved):
Let's see if the image looks more legible.

PR #361, opened 2022-05-26
Add directive "gzip"

Keven-Lin:


nibinqtl (approved):
the default says "expires on;" needs to be fixed.

PR #363, opened 2022-05-26
Fixed broken link and modified cert

marcfishelquantil:
Fixed broken link in the English-language topic Navigating the UI topic. This fix required me to reword the description of Content Management, which now involves purge requests and prefetch requests. For the equivalent Chinese-language topic, I removed the bad link to avoid a 404 error, but did not change the text because the topic has already been translated. Will inform Keven that the content Management subtopic in the Navigating the UI topic needs to be retranslated.


burtonquantil (approved):
ok to merge

PR #366, opened 2022-05-26
High-Res Screens

marcfishelquantil:
Attached are high-resolution screen shots for the Certificates topics. If these look good, I will upload the remaining ones.


nibinqtl (approved):
These pictures are much better!

PR #368, opened 2022-05-31
Remaining High-Res Chinese Screen Shots

marcfishelquantil:
Previously, I uploaded the high-resolution Chinese-language screen shots for the Certificates topics only; those screens were merged with Dev. This upload contains the remaining high-res screen shots for the other Chinese-language topics. I added a few English-languages screen shots as well when the current screen shots were blurry.


burtonquantil (approved):
OK to merge

PR #371, opened 2022-06-02
added slash to links

marcfishelquantil:
While testing the cross-references from my previous commit, I ran across a few that still did not work. Upon closer inspection, I noticed that the leading slash was missing from the cross-reference. For example, the slash before "docs" in the following string (</docs/portal/managing-standby-tasks.md>) was missing and the cross-reference appeared as follows: (<docs/portal/managing-standby-tasks.md>), which caused a 404 error when the link was clicked on docdev.

This has been fixed for all such cross-references in the English- and Chinese-language markdown files. After merging to dev, this should fix all broken cross-references. I will confirm this is so after the merge to dev.


burtonquantil (approved):
OK to merge.

PR #388, opened 2022-07-08
Figures captured with Snipping Tool

marcfishelquantil:


marcfishelquantil:
I have another idea for reducing the size of the graphic. I will try that method in my next Pull Request.

PR #397, opened 2022-08-11
Merge latest changes to production

burtonquantil:
These include the newest release notes and updated CDN Pro API specification.

### DIFF
--- a/en/apispec/openapi.json
+++ b/en/apispec/openapi.json
@@ -697,7 +697,7 @@
           "name": "version",
           "in": "path",
           "required": true,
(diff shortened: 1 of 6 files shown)


marcfishelquantil (approved):
Approved, although I spotted one typo as a remark: "Specifies that a given location can be used for internal requests only. No chang."

burtonquantil:
@marcfishelquantil  i think that's just a Github display issue.
https://docdev.quantil.com/en/cdn/docs/edge-logic/supported-directives#internal has what I believe is the text you mentioned.

"Specifies that a given location can be used for internal requests only. No change to the public version."

PR #407, opened 2022-09-08
Burton ngdoc163

burtonquantil:
I made a few small changes to the text of the document on API rate limiting after re-reading it.
Otherwise, I edited links to refer to appropriate places in our documentation site.

If it looks OK, please approve, and we can check once more on docdev.quantil.com
Someone else would need to do the Chinese translation later.
Thanks


marcfishelquantil (approved):
Approved

PR #409, opened 2022-09-14
Burton ngdoc163

burtonquantil:
Per Bin's suggestion, I made the FAQ a list of questions instead.

I took the liberty of updating the Chinese language FAQ. However, I did not create a file on API rate limiting. This requires someone else to translate.

Please note I changed the title of the API rate limiting FAQ to   "How are CDN Pro API calls rate limited?"  because I did not want our customers to think the topic is about how we might support rate limiting of THEIR API servers accelerated via CDN Pro.


nibinqtl (approved):
lgtm, let's check the effect on docdev

PR #411, opened 2022-09-20
NGDOC-163 #comment the documentation framework doesn't seem to suppor…

burtonquantil:
…t links to internal files unless they are in manifest.json, so let's go back to the original FAQ format.


marcfishelquantil (approved):
Approved.

PR #438, opened 2022-11-10
Burton ncp2603

burtonquantil:
Added description of ENCRYPT_SYMM.
Does this sound OK?
If yes, we'll still need a Chinese translation in the corresponding file.


marcfishelquantil (approved):
Approved

PR #461, opened 2023-01-03
NGDOC-201 Introduce new directives and variables

Jianghai-Quantil:
@nibinqtl pls review


en/docs/edge-logic/built-in-variables.md
```diff
@@ -60,6 +60,8 @@ If you use the CDN Pro portal to edit the property, typing `$` at a position whe
...
 | <span id="served_from_cache"></span><span class="var">$served_from_cache</span> |<span class="badge small" title="numerical value">#</span> Value is 1 for a HIT on edge; 0 for a MISS. | <span class="badge green">Yes</span> | <span class="badge green">Yes</span> |
 | <span id="server_addr"></span><span class="var">$server_addr</span> | IP address of the edge server | <span class="badge yellow">No</span> | <span class="badge green">Yes</span> |
+| <span id="server_level"></span><span class="var">$server_level</span> | <span class="badge small" title="numerical value">#</span> Level of the cache server. 1 means edge, and 2 means intermediate. | <span class="badge yellow">No</span> | <span class="badge green">Yes</span> |
+| <span id="server_name"></span><span class="var">$server_name</span> | Name of the nginx server_name that matches an incoming request. | <span class="badge green">Yes</span> | <span class="badge green">Yes</span> |
```
burtonquantil, line 64: "nginx" should be "Nginx".
I think the text should be "Name of the Nginx server that matches an incoming request."

en/docs/edge-logic/supported-directives.md
```diff
@@ -363,6 +383,16 @@ else { ... }
...
 
+### `ignore_invalid_range`
+
+<span class="badge dark">advanced</span> <span class="badge primary">Proprietary</span>
+
+**Syntax:** `ignore_invalid_range on|off;` <br/>
+**Default:** `ignore_invalid_range off;` <br/>
+**Context:** server, location <br/>
+
+Specifies if invalid range header shall be ignored. When turned on, invalid range header is ignored and 200 full content is returned to client. Otherwise, the client will receive a 416 status code.
```
burtonquantil, line 394: I presume you are referring to https://www.rfc-editr.org/rfc/rfc7233#section-3.1
If so, I suggest replacing range with the capitalized  "Range"
Suggestion:

"Specifies whether to ignore an invalid Range header.  When turned on, an invalid Range header is ignored, and a 200  response with full content is returned to the client. Otherwise, the client will receive a 416 status code. "

en/docs/edge-logic/supported-directives.md
```diff
@@ -981,6 +1021,16 @@ This is an enhanced version of the [open-source version](http://nginx.org/en/doc
...
 
+### `range_reorder`
+
+<span class="badge dark">advanced</span> <span class="badge primary">Proprietary</span>
+
+**Syntax:** `range_reorder on | off [coalescing];` <br/>
+**Default:** `range_reorder off` <br/>
+**Contexts:** server, location
+
+This directive can be used to instruct the proxy server to reorder and merge multiple byte-ranges in one request. When range_reorder is enabled, multiple ranges requested in descending order will be re-ordered to ascending. When range_order is enabled with the coalescing parameter, multiple ranges that are overlapped or are separated by a gap that is smaller than the overhead of sending multiple parts will be coalesced. This ensures that multipart content is returned to client with the 206 status code. When range_reorder is disabled, full content with 200 status code might be returned to client for requests with multiple ranges.
```
burtonquantil, line 1032: change " multiple ranges that are overlapped " to "multiple ranges that overlap"
I also suggest changing "returned to client" with "returned to the client"

en/docs/edge-logic/supported-directives.md
```diff
@@ -1122,6 +1172,18 @@ Sets the size of the slices when fetching large files from the origin. The valid
...
+
+<span class="badge">standard</span> <span class="badge primary">Proprietary</span>
+
+**Syntax:** `slice_verify_header header_name;` <br/>
+**Default:** - <br/>
+**Contexts:** server, location
+
+This directive can be used to specify a header in addition to Etag to check consistency of sliced files. Use this directive in conjunction with[`slice_ignore_etag`](#slice_ignore_etag) if you want to skip Etag and check the specified header only.
+
+The header_name must be of a value other than "etag". The value is case insensitive.
```
burtonquantil, line 1185: I believe "case insensitive" should be hyphenated as "case-insensitive"

burtonquantil (changes requested):
I recommended some changes to improve the English.
Bin needs to confirm whether these variables should all be exposed to customers and check the Chinese translations.

en/docs/edge-logic/built-in-variables.md
Jianghai-Quantil: The way we name variables is confusing. For example, if you look at $server_name and $server_region, you'll find that the term "server" is used to refer to nginx server_name (virtual host) in one variable and to the cache server in another variable. 
I intentionally use "nginx server_name" here, hoping that this variable won't be interpreted as representing the cache server.
Maybe we can improve it to "The Nginx server_name that matches an incoming request."

en/docs/edge-logic/supported-directives.md
Jianghai-Quantil: Updated.

en/docs/edge-logic/supported-directives.md
Jianghai-Quantil: Updated.

en/docs/edge-logic/supported-directives.md
Jianghai-Quantil: Updated.

en/docs/edge-logic/built-in-variables.md
burtonquantil: @Jianghai-Quantil I see.   I had made the suggestion since it is odd to have the word you're trying to explain used in the explanation.

burtonquantil (approved):
OK to merge

nibinqtl (approved):
I made some updates to [Neil-NGDOC-201](https://github.com/mileweb/docs-content-cdn/tree/Neil-NGDOC-201) about the $server_name variable.

PR #471, opened 2023-01-20
updated screen shots

marcfishelquantil:
Burton, this upload should fix the issues with the previous upload and should also address your Skype comments.


burtonquantil (approved):
OK to merge

PR #472, opened 2023-01-20
more adjusted screen shots

marcfishelquantil:
Burton, this batch s/be the last set of screen shots we need. Please approve so we can see the results on docdev.


burtonquantil (approved):
OK to merge

PR #480, opened 2023-02-14
NGDOC-210 Add upstream_origin_only

Jianghai-Quantil:
@nibinqtl @burtonquantil Added a new directive upstream_origin_only. Please help review.

### DIFF
--- a/en/docs/edge-logic/supported-directives.md
+++ b/en/docs/edge-logic/supported-directives.md
@@ -1235,6 +1235,16 @@ Indicates whether to look for each string to replace once or repeatedly. No chan
 
 Enables string replacement in responses with the specified MIME types in addition to “text/html”. No change to the public version.
 
+### `upstream_origin_only`
+
+<span class="badge">standard</span> <span class="badge primary">Proprietary</span>
+
+**Syntax:** `upstream_origin_only on|off;` <br/>
+**Default:** `upstream_origin_only off` <br/>
+**Contexts:** server, location, if in location
+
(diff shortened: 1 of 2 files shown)


en/docs/edge-logic/supported-directives.md
```diff
@@ -1235,6 +1235,16 @@ Indicates whether to look for each string to replace once or repeatedly. No chan
...
 
+### `upstream_origin_only`
+
+<span class="badge">standard</span> <span class="badge primary">Proprietary</span>
+
+**Syntax:** `upstream_origin_only on|off;` <br/>
+**Default:** `upstream_origin_only off` <br/>
+**Contexts:** server, location, if in location
+
+Enables or disables an origin only upstreaming. When set to on, a matching request is forwarded directly to the origin, bypassing any intermediate cache (including [origin shields](/cdn/apidocs#operation/get-cdn-shields)). This can be very useful if you would like some requests to go through intermediate layers while other requests (e.g. auth requests) to be forwarded directly to the origin.
```
burtonquantil, line 1246: I asked Ramprasad to clarify the functionality. Based on my understanding, I would like to modify to describe a bit more:

Enable or disable origin only upstreaming. When set to on, a matching request is forwarded directly to the origin, bypassing any intermediate cache including [origin shields](/cdn/apidocs#operation/get-cdn-shields).  For example, you may have configured your property's directConnection setting to "noDirect" to always use an intermediate cache or "auto" for us to dynamically determine the best course, by default.  The upstream_origin_only directive lets you bypass that setting and allow a subset of requests to go directly to the origin

burtonquantil (changes requested):
I suggest some additional detail to make it clearer to a customer the scenario when they may want to use this new directive.

en/docs/edge-logic/supported-directives.md
Jianghai-Quantil: Suggestion accepted.

burtonquantil (approved):
OK to merge

PR #485, opened 2023-03-07
NGDOC-213 #comment updated release notes and OpenAPI spec to cover th…

burtonquantil:
…e latest release


marcfishelquantil (approved):
You could change "keep track of" to "track, but this is minor. Approved.

PR #511, opened 2023-07-05
NGDOC-227 #comment updated release notes and API spec to cover the la…

burtonquantil:
…test release


marcfishelquantil (approved):
Approved

PR #521, opened 2023-08-18
NGDOC-234 #comment tried to move line to make it look better

burtonquantil:
Fixed the display of the image so it  is next to the associated line of text.
I tested locally.


chiraags (approved):
Good to merge.

PR #524, opened 2023-08-30
NGDOC-232 fix broken links and a missing column delimiter

Jianghai-Quantil:
@burtonquantil Just found 2 issues with the documentation. Please review the revisions.


burtonquantil (dismissed):
@Jianghai-Quantil  Can you also fix the following in the FAQ?
" by assigning value to this variable"  should be "by assigning a value to this variable"?

I also noticed that you added 3 FAQs.  I think "Assign codes to content" and "Use content codes for multiple purposes"
should be subtopics under the "How to classify traffic in reports based on custom rules" section. 
This would be similar to how there are two subtopics under 
https://docdev.quantil.com/en/cdn/docs/edge-logic/faq#what-about-dynamic-content

Jianghai-Quantil:
> @Jianghai-Quantil Can you also fix the following in the FAQ? " by assigning value to this variable" should be "by assigning a value to this variable"?
> 
> I also noticed that you added 3 FAQs. I think "Assign codes to content" and "Use content codes for multiple purposes" should be subtopics under the "How to classify traffic in reports based on custom rules" section. This would be similar to how there are two subtopics under https://docdev.quantil.com/en/cdn/docs/edge-logic/faq#what-about-dynamic-content

They are subtopics in my source file. It is just the tool used for the documentation site does not render the format perfectly. I can change the format as recommended, but there are sub-subtopics which I don't know what format to use. 

If you look at this topic https://docdev.quantil.com/en/cdn/docs/edge-logic/faq#china-delivery-with-and-without-icp-beian, you'll find that the 2 subtopics look like 2 FAQs on the menu, but they are not.

burtonquantil (approved):
OK to merge.
We can revisit the question of the subtopics appearing on the right  menu later.
It's not a blocker.

PR #527, opened 2023-09-02
To enable syntax highlight of nginx code

nibinqtl:


Jianghai-Quantil (dismissed):
Looks good to me. But shall we also make the change for Chinese version?

nibinqtl:
> Looks good to me. But shall we also make the change for Chinese version?

Good point. Just updated the ZH version.

Jianghai-Quantil (approved):
OK to merge.

PR #540, opened 2023-10-13
Update supported-directives.md

nibinqtl:
Fixed wrong information about dropping query string to origin.

### DIFF
--- a/en/docs/edge-logic/supported-directives.md
+++ b/en/docs/edge-logic/supported-directives.md
@@ -534,7 +534,7 @@ This directive specifies the origin from which to fetch the content. It is a wra
 ```nginx
 # when URI is not specified, 
 # origin_pass my_origin; # URI-escaped $uri and the query string will be appended by default
-rewrite ^(.*) $1? break; # the question mark at the end prevents query strings being appended
+rewrite ^(.*) $1? break; # the question mark at the end prevents query strings from being appended
 origin_pass my_origin; # no query string will be forwarded to the origin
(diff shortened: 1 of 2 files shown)


en/docs/edge-logic/supported-directives.md
```diff
@@ -530,17 +530,15 @@ This is a wrapper of the [proxy_limit_rate](http://nginx.org/en/docs/http/ngx_ht
...
 ```nginx
-# when URI is not specified, URL-encoded $uri and query string will be appended by default
-origin_pass my_origin;
-eval_func $uri_uenc URL_ENCODE $uri;
-origin_pass my_origin$uri_uenc$is_args$args; #same as above
-origin_pass my_origin$uri_uenc; #to drop the query string
-origin_pass my_origin/abc$uri_uenc;
+# when URI is not specified, 
+# origin_pass my_origin; # URI-escaped $uri and the query string will be appended by default
+rewrite ^(.*) $1? break; # the question mark at the end prevents query strings being appended
```
burtonquantil, line 537: I would add the word "from" before "being appended".

burtonquantil (approved):
OK to merge. I made one wording suggestion.

en/docs/edge-logic/supported-directives.md
nibinqtl: You can update the branch directly to save another round of review.

burtonquantil (approved):
OK to merge

PR #547, opened 2023-10-26
NGDOC-248 #comment added that we allow an Nginx variable in the Host …

burtonquantil:
…Header field of a property definition


nibinqtl (approved):
look OK

PR #552, opened 2023-11-09
NGDOC-250 #comment fixed issues reported in NGDOC-250 and NGDOC-251

burtonquantil:
also fixed missing items in the Chinese help


marcfishelquantil (approved):
Approved

PR #560, opened 2023-12-05
NCP-3531 Updated the proxy_pass document to mention URL_ESCAPE, and origin client cert in recipe

nibinqtl:


Jianghai-Quantil (approved):
LGTM

PR #566, opened 2023-12-13
Revise description for $request_time

Jianghai-Quantil:
@nibinqtl Please review the change.
Ricky from Apple questioned the meaning of $request_time as described in our public documentation. He said that the current description is not clear as to when the timer stops and what duration the time value represents. Ricky, Jiali and Yanming, they even had a meeting yesterday to discuss this. They want us to revise the description of this variable to better explain the meaning. 

I checked the official nginx documenation and find this:
$request_time: request processing time in seconds with a milliseconds resolution; time elapsed between the first bytes were read from the client and the log write after the last bytes were sent to the client

So I propose using this offical nginx description instead. The current description of this variable on our public documentation is actually an excerpt of the above text from nginx. Not sure if the last few words "and the log write after the last bytes were sent to the client" were deliberately removed from our documentation.


nibinqtl (changes requested):
The document is correct. Nginx official documentation only applies to writing this variable to access log.

Jianghai-Quantil:
@nibinqtl Please review again.

nibinqtl (changes requested):
Since the unit of $request_time is seconds, you should mention the "second" (instead of the millisecond) version of the other variables.

Jianghai-Quantil:
@nibinqtl I misunderstood the part about milliseconds resolution. Just make a change. Please check.

nibinqtl (approved):
I made some changes to the EN version.

PR #570, opened 2023-12-16
modify-notes-for-secret-update

Jianghai-Quantil:
@burtonquantil Please help review. This is a change requested by Wangsu colleagues. The note for secret update on portal documentation is not clear enough, because it doesn't mention what happens when secret value is updated. I simply replace the content with text copied from API documentation.


burtonquantil (approved):
OK to merge

PR #571, opened 2023-12-28
NCP-3510 #comment updated text to refer to the new API. Copied Chines…

burtonquantil:
…e text from Neil in https://github.com/mileweb/docs-content-cdn/compare/dev...ncp-3510


marcfishelquantil (approved):
Approved

PR #589, opened 2024-03-21
NCAS-4756 remove origin_selection_algorithm

Jianghai-Quantil:
@nibinqtl pls review


nibinqtl (approved):
We need to add document to explain the new way to configure this for origins.

Jianghai-Quantil:
Sure. Will need to wait for implementation of related NCAS and NCP tasks before updating public documentation.

PR #610, opened 2024-04-24
merge the latest to production including release notes and spec updates

burtonquantil:


en/docs/edge-logic/paths-to-origins.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+### Fast Route to Origin
+
+There is still one remaining issue for both dynamic and static content: What if the connectivity between the CDN Pro server and the origin is not stable?
+When internet congestions and outages happen, we want to minimize their impact on our business. Here is where the "Fast Route to Origin"
+feature can help greatly. When you enable it in the edge logic with the [`origin_fast_route`](</docs/edge-logic/supported-directives.md#origin_fast_route>) directive, the data to and from the origin will be 
+accelerated by our high-speed data transmission (HDT) platform. This service features our proprietary protocol based on UDP and smart routing technology to ensure stable layer-4 performance regardless of any fluctuation in the public internet.
+
+Please contact our customer support if you need to try this service. Due to the higher cost of HDT, the traffic transmitted through this channel is charged seperately from the edge delivery traffic.
```
marcfishelquantil, line 67: Please change "seperately" to "separately".

zh/docs/edge-logic/paths-to-origins.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+### 快速回源路径
+
+There is still one remaining issue for both dynamic and static content: What if the connectivity between the CDN Pro server and the origin is not stable?
+When internet congestions and outages happen, we want to minimize their impact on our business. Here is where the "Fast Route to Origin"
+feature can help greatly. When you enable it in the edge logic with the [`origin_fast_route`](</docs/edge-logic/supported-directives.md#origin_fast_route>) directive, the data to and from the origin will be 
+accelerated by our high-speed data transmission (HDT) platform. This service features our proprietary protocol based on UDP and smart routing technology to ensure stable layer-4 performance regardless of any fluctuation in the public internet.
+
+Please contact our customer support if you need to try this service. Due to the higher cost of HDT, the traffic transmitted through this channel is charged seperately from the edge delivery traffic.
```
marcfishelquantil, line 66: Please change "seperately" to "separately"

marcfishelquantil (commented):
Please see my three comments.

PR #622, opened 2024-05-21
Emphasize a few directives and variables with backticks

nibinqtl:


en/docs/edge-logic/faq.md
```diff
@@ -53,7 +53,7 @@ If you want to keep any previously assigned value, you can append to this variab
...
... [GraphQL](https://en.wikipedia.org/wiki/GraphQL). In these cases, the POST requests are idempotent and safe as the GET requests, and the response are well cacheable. The only question is how to include the parameters in the request body to the cache key. CDN Pro created the proprietary directive [`proxy_request_body_in_cache_key`](/docs/edge-logic/supported-directives.md#proxy_request_body_in_cache_key) for this exact purpose. When this feature is turned on, our server would calculate an MD5 hash from the request body and append it to the cache key. Due to performance considerations, this only happens when the body size is less than 4kB, otherwise the cache key is not appended and the variable `$ignored_body_in_cache_key` is set to 1 to indicate this fact. You can use this variable with [`proxy_cache_bypass`](/docs/edge-logic/supported-directives.md#proxy_cache_bypass) to avoid serving incorrect cached content. If you really need to include larger request body in the cache key, you are advised to calculate a hash of the request body in the client and pass it through the request header. You can then use the method introduced earlier to include it in the cache key via `$cache_misc`.
```
burtonquantil, line 56: Suggestions:
Change:  "the response are well cacheable" to "the responses are well cacheable".     (Earlier, you referenced the plural "POST requests", so we need to be consistent.

Change: "our server would calculate an MD5 hash from the request body and append it to the cache key" to 
"our server calculates an MD5 hash from the request body and appends it to the cache key"

Change:  " when the body size is less than 4kB, otherwise the cache key"   to split into two sentences:
"when the body size is less than 4kB. Otherwise, the cache key"

Change:  " If you really need to include larger request body" to 
"If you really need to include a larger request body"

burtonquantil (changes requested):
I made a few  recommendations.

en/docs/edge-logic/faq.md
nibinqtl: updated and broke the long line

burtonquantil (approved):
I'd add a comma after "Otherwise",  but it's OK to merge

PR #643, opened 2024-08-05
ngdoc-281 fix wrong directConnection field name

Jianghai-Quantil:
@nibinqtl @burtonquantil I find that we've mistakenly written directConnection as directConnect. Started a PR to fix it.


burtonquantil (approved):
The English looks OK to me

PR #644, opened 2024-08-05
NGDOC-283 #comment updated release notes and API spec to reflect late…

burtonquantil:
…st CDN Pro release


chiraags (approved):
Good to merge.

PR #680, opened 2024-12-27
NGDOC-304

Jianghai-Quantil:
https://quantil.atlassian.net/browse/NGDOC-304


en/docs/edge-logic/faq.md
```diff
@@ -32,6 +32,12 @@ proxy_ignore_cache_control no-cache no-store;
 * The CDN Pro proprietary directive [`proxy_cache_min_age`](</docs/edge-logic/supported-directives.md#proxy_cache_min_age>) can be used to override the `max-age` in the `Cache-Control` header field to enforce a minimum cache time.
 * If you don't want a request to be served from the cache, you can use the [`proxy_cache_bypass`](</docs/edge-logic/supported-directives.md#proxy_cache_bypass>) directive. [`proxy_no_cache`](</docs/edge-logic/supported-directives.md#proxy_no_cache>) can be used to prevent a response from being cached.
 
+Note: Requests that meet any of the following conditions will not be cached, irrespective of edgeLogic and instructions from origin:
```
burtonquantil, line 35: Elsewhere, we refer to " Edge Logic", not "edgeLogic" unless you intend to refer specifically to the edgeLogic field in the create/update property version APIs.

en/docs/edge-logic/supported-directives.md
```diff
@@ -606,15 +615,7 @@ For example:
...
-```nginx
-proxy_no_cache 1;      # do not cache
-proxy_cache_bypass 1;
-# pass the If-Modified-Since field from client to the origin
-origin_set_header If-Modified-Since $http_if_modified_since flag=any;
-origin_pass My-Dynamic-Origin;
-```
-Note that the `flag=any` parameter is necessary in this case. Otherwise the `If-Modified-Since` header won't appear in the requests to the parent servers, hence won't be forwarded to the origin as desired.
+3. The edge servers forward most client request header fields to the parent servers and the origin, except for these ones: `If-Modified-Since`, `If-Unmodified-Since`, `If-None-Match`, `If-Match`, `Range`, and `If-Range`. For cacheable requests, the servers will automatically regenerate these fields based on the cache policy when fetching from the origin. For example, when [slicing](#slice) is enabled, the servers will automatically generate the `Range` header based on size of slices. Thus, the directive shall not be used to set the `Range` header to origin. For non-cacheable requests, those client request headers will still be forwarded to upstream.
```
burtonquantil, line 618: The word "ones" is unnecessary. You can just say "... except for these:"
Should the last sentence end in "forwarded to the upstream server"  ?

en/docs/edge-logic/supported-directives.md
```diff
@@ -732,7 +733,7 @@ If there is no suffix in the time, the configured value is considered in seconds
 **Default:** `proxy_cache_use_stale error timeout;` <br/>
 **Context:** server, location
 
-Determines in which cases a stale cached response can be used during communication with the proxied server. No change to the [public version](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_use_stale). Based on the default setting, the CDN Pro edge server would return stale cached content if there is any problem establishing connection to the origin.
+Determines in which cases a stale cached response can be used during communication with the upstream server. No change to the [public version](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_use_stale). Based on the default setting, the CDN Pro edge server would return stale cached content if there is any problem establishing connection to the origin.
```
burtonquantil, line 736: I think "establishing a connection" is better than "establishing connection"

burtonquantil (approved):
I made a few suggestions , but otherwise, it looks OK.

Jianghai-Quantil:
@nibinqtl I go ahead merging the PR. Will start new PR to make further changes if necessary.

PR #697, opened 2025-01-31
NGDOC-318 #comment updated API spec to describe more supported variables

burtonquantil:
update the Chinese language versions of the files with the latest changes


marcfishelquantil (approved):
Approved

PR #705, opened 2025-02-26
NGDOC-323 NGDOC-322

Jianghai-Quantil:
@nibinqtl @burtonquantil Please review. Related Jira tasks:
https://quantil.atlassian.net/browse/NGDOC-322
https://quantil.atlassian.net/browse/NGDOC-323

### DIFF
--- a/en/docs/edge-logic/supported-directives.md
+++ b/en/docs/edge-logic/supported-directives.md
@@ -680,7 +680,7 @@ That behavior is controlled by another directive [`proxy_no_cache`](#proxy_no_ca
 **Default:** `proxy_cache_convert_head on;` <br/>
 **Context:** server, location
 
-Enables or disables the conversion of the “HEAD” method to “GET” for caching. When enabled, CDN Pro cache servers will convert “HEAD” method to “GET” when reading from cache or sending upstream requests to origin. If your origin expects HEAD requests and this conversion might cause requests to break, you should disable the conversion. For example, if your origin enforces authentication which takes request method as a factor when computing signature for authentication, the conversion of request method is likely to cause unexpected authentication failures.
+Enables or disables the conversion of the “HEAD” method to “GET” for caching. When enabled, CDN Pro cache servers will convert “HEAD” method to “GET” when reading from cache or sending upstream requests to the origin. If your origin expects HEAD requests and this conversion might cause requests to break, you should disable the conversion. For example, if your origin enforces authentication which takes the request method as a factor when computing the signature for authentication, the conversion of the request method is likely to cause unexpected authentication failures.
 
 ### [`proxy_cache_lock_timeout`](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_lock_timeout)
 
@@ -692,7 +692,7 @@ Enables or disables the conversion of the “HEAD” method to “GET” for cac
 
 Sets a timeout for [`proxy_cache_lock`](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_lock). When `proxy_cache_lock` is enabled, only one request at a time will be allowed to populate a new cache element for the same cache key. Other requests of the same cache element will either wait for a response to appear in the cache or the cache lock for this element to be released, up to the time set by this `proxy_cache_lock_timeout` directive. CDN Pro turns on `proxy_cache_lock` platform wide to better control the traffic to the origin servers. However, since locking will introduce unnecessary latency when most of the contents are not cacheable, we make `proxy_cache_lock_timeout` default to 0. If you know that most of the contents are cacheable, you can set it to some higher value to reduce origin traffic. In the meantime, if you have a way to accurately identify uncacheable content, use `proxy_cache_bypass` and `proxy_no_cache` to skip caching and incur the least latency possible. 
 
-We make a change to the open source version such that requests released upon expiration of lock timeout are still allowed to populate the cache element.
+We make a change to the open source version such that requests released upon expiration of the lock timeout are still allowed to populate the cache element.
 
 ### `proxy_cache_max_stale`
 
@@ -1051,7 +1051,7 @@ Overrides the "Sample Rate" specified for the [Real-Time Log](/docs/portal/edge-
 **Default:** `-` <br/>
 **Contexts:** server, location
 
-Set header and its value which will will be included as request header when sending realtime logs to remote server.
+Sets a header and its value which will will be included as a request header when sending realtime logs to the remote server.
 
 ### [`return`](http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#return)
 
--- a/en/docs/edge-logic/supported-directives.md
+++ b/en/docs/edge-logic/supported-directives.md
@@ -350,7 +350,6 @@ This directive belongs to the nginx [rewrite module](http://nginx.org/en/docs/ht
 
 Enables or disables adding or modifying the “Expires” and “Cache-Control” response header fields. No change to the [public version](http://nginx.org/en/docs/http/ngx_http_headers_module.html#expires). This directive affects only the header fields sent to the client. It does not change the cache time of the content on the server.
 
-
 ### [`gzip`](https://nginx.org/en/docs/http/ngx_http_gzip_module.html#gzip)
 
 <span class="badge">standard</span>
@@ -373,6 +372,16 @@ Enables or disables gzipping of responses. No change to the [public version](htt
 CDN Pro always uses gzip and applies it to the default MIME types above. In addition, compression is activated only when the response body size is greater than 1000 bytes. The default behavior should work well for most users. This directive can be used to enable compression on other types. The search and match are case-insensitive. We improved the public version to support up to 20 wildcards like `text/*` and `*javascript`.
 
 Note: Although it is currently allowed to set different MIME types for gzip and brotli compression by using `gzip_types` and [`brotli_types`](#brotli_types), the types set for the two compression algorithms will be merged in the near future. We recommend setting the same value for the two directives if both gzip and brotli compression are enabled.
+
+### [`http2_max_concurrent_streams`](https://nginx.org/en/docs/http/ngx_http_v2_module.html#http2_max_concurrent_streams)
+
+<span class="badge dark">advanced</span>
+
+**Syntax:** `http2_max_concurrent_streams number;` <br/>
+**Default:** `http2_max_concurrent_streams 32;` <br/>
+**Context:** server, location
+
+Sets the maximum number of concurrent HTTP/2 streams in a connection. No change to the open source version, except that the default value is set to 32.
 
 ### [`if/elseif/else`](http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#if)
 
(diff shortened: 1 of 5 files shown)


en/docs/edge-logic/supported-directives.md
```diff
@@ -121,6 +121,10 @@ auth_request /auth$is_args$args;
 ```
 The request will be granted if the auth response returns a 2xx status code or rejected if the auth response returns a 401 or 403 status code. Other status codes are considered an error and return a 500 "internal error" to the client.
 
+Variables in the subrequest context are handled separately from the main request. This avoids variables in the main request context being polluted.
```
burtonquantil, line 124: what does it mean for variables to be "polluted"?

en/docs/edge-logic/supported-directives.md
```diff
@@ -657,25 +672,15 @@ proxy_cache_bypass $http_pragma    $http_authorization;
...
 <span class="badge">standard</span>
 
-**Syntax:** `proxy_cache_lock_age time;` <br/>
-**Default:** `proxy_cache_lock_age 15s;` <br/>
+**Syntax:** `proxy_cache_convert_head on | off;` <br/>
+**Default:** `proxy_cache_convert_head on;` <br/>
 **Context:** server, location
 
-If the last request passed to the upstream server for populating a new cache element has not completed for the specified time, one more request may be passed to the upstream server. No change to the public version.
+Enables or disables the conversion of the “HEAD” method to “GET” for caching. When enabled, CDN Pro cache servers will convert “HEAD” method to “GET” when reading from cache or sending upstream requests to origin. If your origin expects HEAD requests and this conversion might cause requests to break, you should disable the conversion. For example, if your origin enforces authentication which takes request method as a factor when computing signature for authentication, the conversion of request method is likely to cause unexpected authentication failures.
```
burtonquantil, line 683: >will convert “HEAD” method 
Change to
will convert the "HEAD" method

I suggest adding "the" in front of origin.   (i.e., "sending upstream requests to the origin"

Change "which takes request method"  to "which takes the request method" ,  "when computing signature" to "when computing the signature",  and "conversion of request method" to "conversion of the request method"

en/docs/edge-logic/supported-directives.md
```diff
@@ -685,7 +690,9 @@ If the last request passed to the upstream server for populating a new cache ele
...
+Sets a timeout for [`proxy_cache_lock`](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_lock). When `proxy_cache_lock` is enabled, only one request at a time will be allowed to populate a new cache element for the same cache key. Other requests of the same cache element will either wait for a response to appear in the cache or the cache lock for this element to be released, up to the time set by this `proxy_cache_lock_timeout` directive. CDN Pro turns on `proxy_cache_lock` platform wide to better control the traffic to the origin servers. However, since locking will introduce unnecessary latency when most of the contents are not cacheable, we make `proxy_cache_lock_timeout` default to 0. If you know that most of the contents are cacheable, you can set it to some higher value to reduce origin traffic. In the meantime, if you have a way to accurately identify uncacheable content, use `proxy_cache_bypass` and `proxy_no_cache` to skip caching and incur the least latency possible. 
+
+We make a change to the open source version such that requests released upon expiration of lock timeout are still allowed to populate the cache element.
```
burtonquantil, line 695: I suggest "expiration of the lock timeout"

en/docs/edge-logic/supported-directives.md
```diff
@@ -969,17 +976,17 @@ Sets the text that should be changed in the “Location” and “Refresh” hea
 **Default:** `proxy_request_buffering off` <br/>
 **Context:** server, location
 
-Enables or disables buffering of a client request body. No change to the open source version, except that it is disabled by default. Must be set to "on" to enable [appending request body to cache key](#proxy_request_body_in_cache_key).
+Enables or disables buffering of a client request body. No change to the open source version, except that it is disabled by default. The value of this directive does not have an effect on the behavior of [appending request body to cache key](#proxy_request_body_in_cache_key).
```
burtonquantil, line 979: I suggest "appending the request body to the cache key"

en/docs/edge-logic/supported-directives.md
```diff
@@ -1026,25 +1033,25 @@ This is an enhanced version of the [open-source version](http://nginx.org/en/doc
...
-<span class="badge">standard</span> <span class="badge primary">proprietary</span>
+<span class="badge">advanced</span> <span class="badge primary">proprietary</span>
 
-**Syntax:** `realtime_log_downsample factor;` <br/>
+**Syntax:** `remote_log_set_header name value;` <br/>
 **Default:** `-` <br/>
 **Contexts:** server, location
 
-Overrides the "Sample Rate" specified for the [Real-Time Log](/docs/portal/edge-configurations/creating-property#real-time-log). `factor` can be an integer in [0, 65535] or a variable. A value of 0 disables the logging; 1 means do not downsample; N>1 means one log entry for every N requests. If the variable value is an empty string, the default setting is not overridden. An invalid value in the variable results in a factor of 100. The actual sample factor can be included in the log with the variable [`$realtime_log_ds_factor`](/docs/edge-logic/built-in-variables#realtime_log_ds_factor). This directive can only be used in the Edge Logic.
+Set header and its value which will will be included as request header when sending realtime logs to remote server.
```
burtonquantil, line 1054: There is an extra "will"
I recommend changing to : 
"Sets the header and its value which will be included as the request header when sending realtime logs to the remote server."

burtonquantil (changes requested):
I made a few suggestions to the English text.

en/docs/edge-logic/supported-directives.md
Jianghai-Quantil: Means the value of a variable being unexpectedly overriden. "pollution" is a metaphoric way to describe that.

en/docs/edge-logic/supported-directives.md
Jianghai-Quantil: People reading the document would have at least some knowledge about HTTP and CDN. So they should know what request body and cache key mean. I believe the original text is clear enough. I'd prefer keeping texts concise.

en/docs/edge-logic/supported-directives.md
burtonquantil: I have not seen the word "polluted"  used in that context.   It would be clearer to developers if we said something like "This prevents the variables in the main context from being unexpectedly overwritten."

en/docs/edge-logic/supported-directives.md
burtonquantil: Sorry, my suggestion was not to clarify HTTP and CDN but was about using proper English versus omitting  the articles ("a", "an", and "the")  though people can typically understand the intent without them.

en/docs/edge-logic/supported-directives.md
Jianghai-Quantil: I understand that you always prefer to make texts crystally clear, but it seems to me that you'd often times worry too much about if readers could understand something. I throw the text to ChatGPT, and ChatGPT understands it perfectly.

My question: Does expression like below make sense? {{the expression in discussion}}
Answer from ChatGPT: 
Yes, the expression makes sense, particularly in the context of web server or application server architecture, where there may be a need to handle subrequests independently from the main request. Let's break down the concepts in your statement:

Subrequest Context: This refers to a situation where a request initiated by a main request spawns additional subrequests. Subrequests can be used for various reasons, such as retrieving data from different sources, including additional resources, or performing asynchronous tasks.

Variables in Subrequest Context: These are the variables that are specific to a subrequest. By handling them separately from those in the main request, you ensure that any operations or changes specific to the subrequest do not inadvertently affect the main request's data or flow.

Handling Separately: This approach is necessary to maintain the integrity and state of each request. Isolating the variables prevents unintended side effects, such as data corruption or incorrect logic execution, when the responses are finally compiled to fulfill the main request.

Avoiding Pollution in Main Request Context: "Pollution" here refers to undesired or unintended alterations to the main request's context as a result of changes made within a subrequest. If variables within subrequests were to impact the main request, it could lead to inconsistencies and potential errors in processing the main request.

In web servers like Nginx or application architectures using microservices, managing request contexts separately is vital for reliable and predictable request handling. This isolation helps make complex systems more maintainable and secure, as well as facilitates concurrent request processing.

I also googled expression like "variable pollution". Here are some reference posts on Stackoverflow. 
https://stackoverflow.com/questions/49808936/how-to-prevent-polluting-variables-in-a-jest-test
https://stackoverflow.com/questions/4076799/access-local-variable-in-function-from-module

I'd prefer to keep the metaphoric expression.

en/docs/edge-logic/supported-directives.md
Jianghai-Quantil: But proper English usually means excessive use of words like articles and thus lengthy texts. 
BTW, "appending request body to cache key" have been there in our documentation since quite a while. This is not something I change to this time.

en/docs/edge-logic/supported-directives.md
```diff
@@ -685,7 +690,9 @@ If the last request passed to the upstream server for populating a new cache ele
...
+Sets a timeout for [`proxy_cache_lock`](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_lock). When `proxy_cache_lock` is enabled, only one request at a time will be allowed to populate a new cache element for the same cache key. Other requests of the same cache element will either wait for a response to appear in the cache or the cache lock for this element to be released, up to the time set by this `proxy_cache_lock_timeout` directive. CDN Pro turns on `proxy_cache_lock` platform wide to better control the traffic to the origin servers. However, since locking will introduce unnecessary latency when most of the contents are not cacheable, we make `proxy_cache_lock_timeout` default to 0. If you know that most of the contents are cacheable, you can set it to some higher value to reduce origin traffic. In the meantime, if you have a way to accurately identify uncacheable content, use `proxy_cache_bypass` and `proxy_no_cache` to skip caching and incur the least latency possible. 
+
+We make a change to the open source version such that requests released upon expiration of the lock timeout are still allowed to populate the cache element.
```
burtonquantil, line 695: I suggest a slight change to the beginning of the last sentence: 

We modified the open source version so that requests released  ...

en/docs/edge-logic/supported-directives.md
```diff
@@ -861,7 +868,7 @@ Determines whether the connection with an upstream server should be closed when
 **Default:** `-` <br/>
 **Context:** server, location
 
-Disables processing of certain response header fields in the response from the origin. It is most commonly used to ignore caching instructions such as the `Cache-Control` or `Expires` fields from the origin. No change to the [open-source version](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_ignore_headers). If you need to ignore only some of the `cache-control` directives, use the [`proxy_ignore_cache_control`](#proxy_ignore_cache_control) directive.
+Disables processing of certain response header fields in the response from the origin. It is most commonly used to ignore caching instructions such as the `Cache-Control` and `Expires` fields from the origin. No change to the [open-source version](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_ignore_headers), except that we couple `Cache-Control` and `Expires` together. Configuring this directive to ignore `Cache-Control` will cause `Expires` to be also ignored, or vice verca. If you need to ignore only some of the `cache-control` directives, use the [`proxy_ignore_cache_control`](#proxy_ignore_cache_control) directive.
```
burtonquantil, line 871: "vice verca"  should be "vice versa"

burtonquantil (approved):
I made one new suggestion along with a request to replace "verca"
You'll need Bin to confirm the Chinese text.

en/docs/edge-logic/supported-directives.md
nibinqtl: @Jianghai-Quantil we should still follow the correct English grammar, although it is a bit longer.

nibinqtl (changes requested):

nibinqtl (approved):
Looks OK to me.

PR #709, opened 2025-03-12
ngdoc-327

Jianghai-Quantil:

Note that access_log_downsample is not allowed in location and if in location, because such usage is rejected by edge logic validator. https://quantil.atlassian.net/browse/NCAS-5256?focusedCommentId=1477379
I take the chance to update http2_max_concurrent_streams to remove location from supported contexts, for the same reason.

We can make a second update, if edge logic validator is changed to allow such usage later on.


burtonquantil (approved):
The updates in the English files look OK to me.

nibinqtl (approved):

PR #711, opened 2025-03-13
More brand substitutions

nibinqtl:


nibinqtl:
Replaced all hard-coded brands with variables. Some links to www.quantil.com may be broken, but it is OK.

Jianghai-Quantil (approved):
Look good to me

PR #712, opened 2025-03-13
NGDOC-328 support brand substitute

nibinqtl:


burtonquantil (approved):
OK to merge.
Please note that I will be checking in changes to the original source files from which the openapi.json is generated.

PR #718, opened 2025-03-26
ncp-3779

Jianghai-Quantil:


nibinqtl (approved):
It is OK to remove the text about LB Hash key.
But we really need to get the ZH version translated.

Jianghai-Quantil:
OK. I'll spend some time in Q2 translating the documents.

PR #724, opened 2025-04-14
ncas-5290

Jianghai-Quantil:
mention the default factor for access_log_downsample when variable value is invalid. @nibinqtl @burtonquantil please review.

### DIFF
--- a/en/docs/edge-logic/supported-directives.md
+++ b/en/docs/edge-logic/supported-directives.md
@@ -16,7 +16,7 @@ In the following list, the <span class="badge">standard</span> directives are av
 **Default:** `-` <br/>
 **Contexts:** server
 
-Downsamples the local access log. A `factor` of N means one log entry for every N requests. It can be used to reduce the amount of access log to download from the portal or API. A log field can be defined with the keyword `%samplerate` to show this factor. This directive has no effect on the real-time log, whose downsampling is controlled by [`realtime_log_downsample`](#realtime_log_downsample). We may also use this directive to prevent properties with large request volume from overloading the log processing system.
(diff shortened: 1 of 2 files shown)


en/docs/edge-logic/supported-directives.md
```diff
@@ -16,7 +16,7 @@ In the following list, the <span class="badge">standard</span> directives are av
...
+Downsamples the local access log. A `factor` of N means one log entry for every N requests. The factor specified can be a positive integer or a variable. The value is 1 by default which means no downsapling, if the directive is not configured or when the variable value is invalid. This directive can be used to reduce the amount of access log to download from the portal or API. A log field can be defined with the keyword `%samplerate` to show this factor. This directive has no effect on the real-time log, whose downsampling is controlled by [`realtime_log_downsample`](#realtime_log_downsample). We may also use this directive to prevent properties with large request volume from overloading the log processing system.
```
burtonquantil, line 19: There is a misspelling "downsapling".
I suggest rewriting the sentence as :
"The default value of 1, meaning no downsampling, is used if the directive is not configured or when the variable value is invalid. "

Also, I think "access log" should be "access logs"

burtonquantil (changes requested):
I suggested a couple of changes.

Jianghai-Quantil:
@burtonquantil Please review again.

burtonquantil (dismissed):
My second suggestion 
was actually to use "logs" in " This directive can be used to reduce the amount of access log"

Jianghai-Quantil:
@burtonquantil Please review again. Apply plural form for access log.

burtonquantil (approved):
Looks good

PR #750, opened 2025-07-29
NGDOC-343 #comment updated text and images to fix issues reported by …

burtonquantil:
…tester.
After some consideration, I didn't make use of the two Chinese language AWS recipe images provided by Marc since we have not updated the rest of the images and don't have a Chinese translation of the text on the page.


en/docs/resources/images/certificates/create-cacertificate.png
Jianghai-Quantil, line 1: There is unexpected highlight on the Comment field. The screenshot should be replaced.

burtonquantil: OK. I'll get a new image from Marc tomorrow.

PR #771, opened 2025-09-18
NCAS-5415 #comment updated description of access control rules to ind…

burtonquantil:
…icate wildcard hostname support


marcfishelquantil (approved):
Approved

